PR #7, opened 2015-01-29
[WIP] Downloader slot usage optimization

sibiryakov:
Imagine, we have a queue of 10K urls from many different domains. Our task is to fetch it as fast as possible. At the same time we have a prioritization which tends to group URLs from the same domain. During downloading we want to be polite and limit per host RPS. So, picking just top URLs from the queue leeds us to the time waste, because connection pool of Scrapy downloader most of time underused. 

In this PR, I'm addressing this issue by propagating information about overused hostnames/IPs in downloader pool.


crawlfrontier/contrib/scrapy/schedulers/frontier.py
```diff
@@ -154,4 +154,12 @@ def _get_exception_code(self, exception):
...
     def _request_is_redirected(self, request):
-        return request.meta.get('redirect_times', 0) > 0
\ No newline at end of file
+        return request.meta.get('redirect_times', 0) > 0
+
+    def _get_overused_keys(self):
+        keys = []
+        downloader = self.crawler.engine.downloader
+        for (key, slot) in downloader.slots.iteritems():
+            if len(slot.active) / float(slot.concurrency) > 1.0:
```
shane42: I am not sure this is the right formula to use here.

How about we use a scrapy setting which specifies the maximum queued requests per key (domain/IP) before we consider it overused? I think that it's a good idea to allow some queued requests, since the current requests may finish before the frontier's get_next_requests is called again.

crawlfrontier/contrib/scrapy/schedulers/frontier.py
```diff
@@ -154,4 +154,12 @@ def _get_exception_code(self, exception):
...
 
     def _request_is_redirected(self, request):
-        return request.meta.get('redirect_times', 0) > 0
\ No newline at end of file
+        return request.meta.get('redirect_times', 0) > 0
+
+    def _get_overused_keys(self):
+        keys = []
+        downloader = self.crawler.engine.downloader
+        for (key, slot) in downloader.slots.iteritems():
```
shane42: no need for parenthesis with (key, slot)

crawlfrontier/core/manager.py
```diff
@@ -299,6 +299,8 @@ def get_next_requests(self, max_next_requests=0):
         will be used instead. (:setting:`MAX_NEXT_REQUESTS` setting).
 
         :param int max_next_requests: Maximum number of requests to be returned by this method.
+        :param list overused_keys: Hostnames or ip addresses (depends on Spider settings) having already full queues of
```
shane42: We should mention which settings

shane42:
Looks like a good start to me, I am glad the Scrapy integration turned out to be straightforward.

We'll need a backend that can take advantage of this and docs + tests.

Should the PR be against the development branch @gatufo , is there contributing docs?

crawlfrontier/contrib/scrapy/schedulers/frontier.py
shane42: no need for parenthesis with (key, slot)

crawlfrontier/contrib/scrapy/schedulers/frontier.py
sibiryakov: I like the idea! I was trying to achieve the same goal, by incorporating this formula here. The maximum you mentioned differs from domain to domain. For example, domain x allows us to query for 20 RPS, and domain y - 0.5 RPS. Let's assume, that we want to refine the overused slots often, and aiming for efficient download. So we want, get_next_requests to be called once per 10 secs. Therefore, maximum value for domain x would be like 200 (20 times 10), and for domain y - 5. A huge difference.

So, I think such (or similar) formula will work better here.

sibiryakov:
@shane42, I have such backend. It's based on ranking by score, which is calculated in spider. I think I'll contribute it also.

crawlfrontier/contrib/scrapy/schedulers/frontier.py
sibiryakov: Added option "OVERUSED_SLOT_FACTOR"

crawlfrontier/core/manager.py
sibiryakov: done.

gatufo:
@shane42, not contributing docs yet, I'll work on it.

sibiryakov:
Guys, please check again. If it's almost ready.

crawlfrontier/core/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        for request in self._pending_requests:
+            key = self._get_key(request, overused_keys.type)
+            if key not in overused_keys:
+                requests.append(request)
+                _remove.append(request)
+            if len(requests) == max_n_requests:
+                break
+
+        for r in _remove:
+            self._pending_requests.remove(r)
```
kmike: `self._pending_requests.difference_update(_remove)` could be a bit better

crawlfrontier/core/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def get_next_requests(self, max_n_requests, overused_keys):
+        if self._log:
+            self._log("Overused keys: %s" % str(overused_keys))
+            self._log("Pending: %i" % len(self._pending_requests))
+        requests = []
+        _remove = []
+        for request in self._pending_requests:
+            key = self._get_key(request, overused_keys.type)
+            if key not in overused_keys:
```
kmike: is this O(N) lookup?

crawlfrontier/core/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for r in _remove:
+            self._pending_requests.remove(r)
+
+        if len(requests) == max_n_requests:
+            return requests
+
+        for request in self._get(max_n_requests-len(requests), overused_keys):
+            key = self._get_key(request, overused_keys.type)
+            if key in overused_keys:
```
kmike: O(N) lookup?

crawlfrontier/utils/managers.py
```diff
@@ -27,8 +27,8 @@ def add_seeds(self, seeds):
         frontier_seeds = [self.request_converter.to_frontier(seed) for seed in seeds]
         self.manager.add_seeds(seeds=frontier_seeds)
 
-    def get_next_requests(self, max_next_requests=0):
-        frontier_requests = self.manager.get_next_requests(max_next_requests=max_next_requests)
+    def get_next_requests(self, max_next_requests=0, overused_keys=[]):
```
kmike: why isn't it a custom list subclass here?

examples/scripts/07_frontier_from_settings.py
```diff
@@ -12,7 +12,7 @@
 
 if __name__ == '__main__':
     # Create graph
-    graph = graphs.Manager('sqlite:///data/graph.db')
+    graph = graphs.Manager('sqlite:///graph_2014_08_25_wo_pdf2.db')
```
kmike: it looks rather specific :) What's thereason for this change?

crawlfrontier/core/manager.py
```diff
@@ -291,7 +291,7 @@ def add_seeds(self, seeds):
                                  obj=seeds,
                                  return_classes=(list,))  # TODO: Dar vuelta
 
-    def get_next_requests(self, max_next_requests=0):
+    def get_next_requests(self, max_next_requests=0, overused_keys=OverusedKeys()):
```
kmike: I see that the code carefully avoids mutating overused_keys variable, so using mutable default doesn't cause a bug here. But it could be better to set it to None and then set to `OverusedKeys()` in the function body, just to be on a safe side.

docs/source/topics/frontier-settings.rst
```diff
@@ -202,6 +202,15 @@ Default: ``crawlfrontier.utils.fingerprint.sha1``
...
 
+.. setting:: OVERUSED_SLOT_FACTOR
+
+OVERUSED_SLOT_FACTOR
+------------------------
+
+Default: ``1.0``
+
+This is an amount of overall active connections to concurrent connections allowed in slot(per-domain/ip) after which slot
+ is considered overused.
```
kmike: Can len(active) be greater than concurrency? I think it makes sense to note what the default argument means. If I'm not mistaken, now it means "never consider slots overused" because the inequality is strict, and it seems a number of active requests can't exceed the concurrency number. But maybe it can ;) and in this case the meaning of the default value is different.

kmike:
The approach looks good.

docs/source/topics/frontier-settings.rst
sibiryakov: It's a tricky topic. It can be greater and and it can be lower, it all depends on what URL's spider sent there at some moment in time. Concurrency can be affected by CONCURRENT_REQUESTS_PER_IP/DOMAIN. Number of active requests in the slot are influenced by delay (which is adaptive with AUTOTHROTTLE), remote web resource request processing time, or even how overloaded with parsing the whole crawler process. The default meaning is "consider slot overused when count of active connections is more than concurrent allowed for that slot". If you have concurrent=10, and active=11 slot is overused.

crawlfrontier/core/__init__.py
sibiryakov: yes, it is.

crawlfrontier/core/__init__.py
sibiryakov: Aha. It needs some more planning on dropping logic. It could consume too much RAM.

sibiryakov:
I also slightly decompose the Request/Response objects creation in SQLiteBackend (sqlalchemy.py), to make it more suitable for extension.

gatufo:
I like the idea, but I'm not sure about the way of communicate the overused keys. I think we should keep frontier API as simple as possible. 

You're getting domain/ip overused info from scrapy downloader, frontier knows about crawling results throught `page_crawled` method. Is it possible to move this logic to the backend somehow and make this feature also available for non-scrapy crawlfrontier users?

sibiryakov:
Overused domain/ip isn't connected with what was crawled. It's about how responsive particular website/network to scrapy downloader. If downloading is slow, we need to know about that in get_next_requests()

sibiryakov:
Guys @shane42 @kmike @gatufo please have a look at this PR again. My goal for the changes were to make it more clear what it does, and how to use it outside of Scrapy. I would like to point your attention to "Efficient parallel downloading" article in documentation and new example with grequests.

gatufo:
I see overused_keys is now packed into downloader_info, but still not sure about api changes.
@shane42, @kmike what about adding some notification method to the frontier? like:

    def send_message(self, message, **kwargs):
        pass

that could be used to send this downloader_info:

    frontier.send_message('downloader_info', downloader_info=downloader_info)

sibiryakov:
@gatufo In the context of overfilled slots of downloader queue the downloader_info should contain *most up-to-date* information. Because situation in downloader, when it fetches from 100 connections in parallel (and up to 10 from the same domain) is changing very quickly. From that point it makes sense to send it with the get_next_requests() call. To avoid mess.

May be there are other reasons why notifications in the form of frontier.send_message() can be useful, but at the moment I do not see them.

kmike:
What do you think about a more loose approach - add an optional "options" argument to `get_next_requests` method? It will be a dict without a strict structure (no need to subclass DownloaderInfo in order to add new info). We'll document that some backends support 'overused_keys' key inside this dict.

A general notifications method requires the frontier to store more data, which can be good or bad.

I'm not sure I understand the big picture good enough, but with `.send_message()` there could be a race condition if it is async - it could be the case we call `send_message` with the new options and then `get_new_requests`, but `get_new_requests` haven't seen new options yet. So passing options in a "transactions", in `get_new_requests` method itself sounds good for me.

gatufo:
I'm with @kmike on this, adding a kwargs argument to frontier methods will allow to easily add new functionality to the frontier without creating specific feature parameters to the API.

We recently faced the same problem with hcf backend and solved it in the same way: https://github.com/scrapinghub/crawl-frontier/commit/33ad7b494b035574068bdee1c19ded598724d895

sibiryakov:
We could end-up that someone will be sending the Scrapy engine object through that parameters to the backend. Too much freedom can make the code terrible. But, if it solves the problem of propagation downloader state to the CF backend I can live with that ;)

sibiryakov:
@gatufo and @kmike Please have a look!

sibiryakov:
Closed due https://github.com/scrapinghub/crawl-frontier/pull/21

PR #9, opened 2015-02-18
Avoid tldextract issuing an HTTP request and use include TLD snapshot

bertinatto:
Avoid junk messages when using Crawl Frontier as an egg.


sibiryakov:
@gatufo May be it's also worth merging it to development branch?

gatufo:
I'll do it soon, I want to finish some features for this backend first. Won't take long ;-)

PR #17, opened 2015-03-16
Mongo db backend updated

sibiryakov:
Based on a solution made by @bbotella at https://github.com/scrapinghub/crawl-frontier/pull/6


sibiryakov:
@plafl @kalessin please have a look.

plafl:
Looks OK to me

sibiryakov:
Great, I think I'll do some docs and tests.

sibiryakov:
@bbotella Honestly, I don't think people will be using MongoDB backend because of performance, instead because of ease of development and schemaless concept. Therefore, I would stay with most transparent MongoDb item possible.

PR #18, opened 2015-03-17
avoid downloader to become idle in order to take new requests from backend, in order to improve performance.

kalessin:
@sibiryakov can please review/merge


sibiryakov:
Actually everything is fine. May be 
```python
len(self) < self.crawler.engine.downloader.total_concurrency
```
would be more readable?

kalessin:
@sibiryakov you are right. Just updated PR with your observation

PR #20, opened 2015-03-18
usage of tldextract is responsible of a significative performance decrease

kalessin:
@sibiryakov  as you suggested I have profiled jobs and process spents an important amount of time (near 30-40% of total) in cPickle.load() method, which is used by tldextract. Disabling DomainMiddleware jobs works much faster, and are not needed by any CF component. So this middleware should be enabled only when required.

### DIFF
--- a/crawlfrontier/settings/default_settings.py
+++ b/crawlfrontier/settings/default_settings.py
@@ -6,9 +6,7 @@ import logging
 REQUEST_MODEL = 'crawlfrontier.core.models.Request'
 RESPONSE_MODEL = 'crawlfrontier.core.models.Response'
 MIDDLEWARES = [
-    'crawlfrontier.contrib.middlewares.domain.DomainMiddleware',
     'crawlfrontier.contrib.middlewares.fingerprint.UrlFingerprintMiddleware',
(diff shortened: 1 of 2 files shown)


crawlfrontier/settings/default_settings.py
```diff
@@ -6,9 +6,9 @@
 REQUEST_MODEL = 'crawlfrontier.core.models.Request'
 RESPONSE_MODEL = 'crawlfrontier.core.models.Response'
 MIDDLEWARES = [
-    'crawlfrontier.contrib.middlewares.domain.DomainMiddleware',
+    # 'crawlfrontier.contrib.middlewares.domain.DomainMiddleware',
```
sibiryakov: IMO, it doesn't make sense to leave comments there. One could easily find them in docs or in sources. So please feel free to remove them.

kalessin: ok. wasn't sure to remove, but I like more to remove :)

sibiryakov:
It would be great to have a tests checking that all components in request.meta that are required for basic CF functionality are added by middlewares. So removing some important middleware or braking it would cause test to fail.

sibiryakov:
Please also update the docs for frontier settings.

sibiryakov:
Overall finding is great!

PR #22, opened 2015-03-23
Fixing hitting get_next_requests on each attempt to get buffered request

sibiryakov:
When it's not possible to fully fill the downloader queue, get_next_requests is called on each scheduler's attempt to get new request in Scrapy Engine. Resulting to blocking the event loop and timing out of requests that are already in the queue.

The solution in this PR introduces 30 seconds delay when backend is returning zero results, allowing rest of the schedulers buffer to exhaust, event loop to unblock and continue processing the queue.

### DIFF
--- a/crawlfrontier/contrib/scrapy/schedulers/frontier.py
+++ b/crawlfrontier/contrib/scrapy/schedulers/frontier.py
@@ -75,12 +75,14 @@ class CrawlFrontierScheduler(Scheduler):
         self.stats_manager = StatsManager(crawler.stats)
         self._pending_requests = deque()
         self.redirect_enabled = crawler.settings.get('REDIRECT_ENABLED')
-        self._delay_next_call = 0.0
 
         frontier_settings = crawler.settings.get('FRONTIER_SETTINGS', None)
         if not frontier_settings:
             log.msg('FRONTIER_SETTINGS not found! Using default frontier settings...', log.WARNING)
         self.frontier = ScrapyFrontierManager(frontier_settings)
+
+        self._delay_on_empty = self.frontier.manager.settings.get('DELAY_ON_EMPTY')
+        self._delay_next_call = 0.0
 
     @classmethod
     def from_crawler(cls, crawler):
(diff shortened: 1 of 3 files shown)


sibiryakov:
@kalessin please have a look!

crawlfrontier/contrib/scrapy/schedulers/frontier.py
```diff
@@ -136,10 +138,18 @@ def has_pending_requests(self):
...
             info = self._get_downloader_info()
-            for request in self.frontier.get_next_requests(key_type=info['key_type'], overused_keys=info['overused_keys']):
+            requests = self.frontier.get_next_requests(key_type=info['key_type'], overused_keys=info['overused_keys'])
+            for request in requests:
                 self._add_pending_request(request)
+
+            # If we got empty requests list, it means backend has nothing to return. Delay will help us to exhaust 
+            # the rest of the buffer and inform the code upper on the stack, that current buffer is empty, without 
+            # hitting backend on every request.
+            self._delay_next_call = clock() + 30.0 if not requests else 0.0
```
kalessin: When not requests, self._delay_next_call is assigned 0.0. Is that what you want? from the reading of the logic seems that what you should do is to make, in this case, self._delay_next_call = clock(), so the correct line should be:

self._delay_next_call = clock() + (30.0 if not requests else 0.0)

right?

sibiryakov: The desired behavior is if we got zero requests from backend, then delay the next call to backend. If not, then no delay is needed. So self._delay_next_call = clock() or 0.0 doesn't change the behavior. Assigning it to 0.0 is just cheaper (no need for system call).

sibiryakov:
Related to https://github.com/scrapinghub/crawl-frontier/pull/19

kalessin:
Ok, looks good to me

kalessin:
just is a bit weird to me that first time self._delay_next_call is initialized to clock() and then to 0.0. Logic is not broken but it is a bit confusing.

vshlapakov:
It looks good for me too, except a tiny moment: probably we don't want hard-coded 30 seconds delay for any backend, it's too restrictive in common sense, so I'd move it to the frontier settings to make it more flexible (with default value).

kalessin:
+1

sibiryakov:
@kalessin, @vshlapakov please check!

vshlapakov:
Thanks, I'm fine with the patch.

sibiryakov:
@vshlapakov Could you merge it?

PR #23, opened 2015-03-26
[WIP] Canonical url solver architecture

sibiryakov:
unnamed

Each document in the web is identified with it's own checksum. At the same time, each document can have few different URLs (even on different domains) and graph of nodes redirecting to it. If our crawler will find the same document starting from some of redirect nodes (URLr1) or using one of it's own URLs, will system detect that this document is already known? What URL we will assign to the document at the end? From my experience, chances that we will face with document URL ambiguity are quite big, since redirects are quite common, especially when dealing with URL shorteners.

Currently Crawl Frontier has no such capability, and it's architecture isn't reflecting this issue. This PR offers the solution in form of new type of backend, called CanonicalSolver. Main purpose for it is to provide any component in CF with canonical url of the document. Integration in crawl frontier internal pipeline allows to hook on all types of events in CF. CanonicalSolver can have it's own persistent storage with redirect graph, and use document content for identification purposes.


vshlapakov:
Actually I'm fine with the idea of expanding the logic for redirect urls, as well as with the current implementation. Long chains of redirects for an arbitrary url is one of the first issues that we've seen in the beginning of `crawl-frontier` project. 
At the time it was solved with similar way, by getting 1st url from the chain if such, as it was done [here](https://github.com/scrapinghub/crawl-frontier/commit/13fcff7d8ed8413a6f9d4c0cfaf89bea983527d1#diff-dd3e6c1820098a29d225be87a8955322R66), but for some reasons (probably nobody used redirects at that moment, or it was removed for simplicity) I can't see it in the code right now.
This approach will work for most cases, but in the case of shorteners for example - there will be several different urls pointed to the same real page. In such way it can affect weight of the real url for some backends (like `OPIC`), and substantially change crawl results. 
In other words, it's the complex issue, and there's no silver bullet to solve it at once, but we definitely should continue to move in this direction, as redirects are everywhere in the web, and it's a core feature of `crawl-frontier`.
The issue is also related with NDD, but I don't think that NDD could/should be solved as a part of crawl-frontier: it's a separate heavy-weight task, and much more coresponds to post-processing step.
About PR, it touches some basic classes, and I'd like to hear @gatufo's opinion as a maintainer of `crawl-frontier`, if he's fine with it, I'll merge it at once.

crawlfrontier/contrib/canonicalsolvers/basic.py
```diff
@@ -0,0 +1,44 @@
...
+        pass
+
+    def get_canonical_url(self, obj):
+        canonical_url = obj.url
+        canonical_fingerprint = obj.meta['fingerprint']
+        canonical_domain_fingerprint = obj.meta['domain']['fingerprint'] if 'domain' in obj.meta else str()
+        if 'redirect_urls' in obj.meta:
+            redirect_urls = obj.meta['redirect_urls']
+            redirect_fingerprints = obj.meta['redirect_fingerprints']
+            redirect_domains = obj.meta['redirect_domains']
```
vshlapakov: Why do we need to track redirect domains chain? What is the use of it?

crawlfrontier/core/components.py
```diff
@@ -107,3 +107,19 @@ class Middleware(Component):
     """Interface definition for a Frontier Middlewares"""
     __metaclass__ = ABCMeta
     component_name = 'Base Middleware'
+
+
+class CanonicalSolver(Component):
+    """Interface definition for a Frontier Canonical Solver"""
+    __metaclass__ = ABCMeta
+    component_name = 'Base CacnonicalSolver'
```
vshlapakov: Small typo error

crawlfrontier/contrib/canonicalsolvers/basic.py
sibiryakov: Well, this decision is full of trade-offs. This data is needed in backend because fingerprint, url and domain is used for record identification purposes and batch generation. The fingerprint and domain related logic is managed and configured in CF middlewares, so I didn't want to overengineer this dealing with settings in CanonicalSolver, etc. At the end of the day we save some CPU cycles caching those values.

sibiryakov: From private chat: NDD is near duplicate detection.

crawlfrontier/core/components.py
sibiryakov: Please point me to the typo, I don't see it ;)

redapple:
```
CacnonicalSolver
  ^
```

crawlfrontier/contrib/canonicalsolvers/basic.py
vshlapakov: Ah, I see, it's more like a list of triples `[(url, fprint, domain),]` rather than 3 lists of values, you're using it to skip extracting a domain and getting a fingerprint for `page_crawled` event (and for `request_error`). 
Ok, I'm just worrying if we really need to get all these values for each request url in a redirects chain, usually we need only one of them and it can be overkill, as `tldextract` is not very fast.
Probably it makes sense to keep only the first and the last, and a chain length, but it's more like a premature optimization for me now, I propose to follow "make it work -> make it right -> make it fast" and now we're somewhere near 2nd point :)

PR #25, opened 2015-04-06
[WIP] Canonical url solver architecture

sibiryakov:
Refined version.
Please see full description and discussion at https://github.com/scrapinghub/crawl-frontier/pull/23

### DIFF
--- a/crawlfrontier/contrib/backends/sqlalchemy/__init__.py
+++ b/crawlfrontier/contrib/backends/sqlalchemy/__init__.py
@@ -122,7 +122,7 @@ class SQLiteBackend(Backend):
 
     def add_seeds(self, seeds):
         for seed in seeds:
-            url, fingerprint = self.manager.canonicalsolver.get_canonical_url(seed)
+            url, fingerprint, _ = self.manager.canonicalsolver.get_canonical_url(seed)
             db_page, _ = self._get_or_create_db_page(url=url, fingerprint=fingerprint,
                                                      request_or_response=seed)
         self.session.commit()
@@ -142,7 +142,7 @@ class SQLiteBackend(Backend):
         return next_pages
 
     def page_crawled(self, response, links):
-        url, fingerprint = self.manager.canonicalsolver.get_canonical_url(response)
+        url, fingerprint, _ = self.manager.canonicalsolver.get_canonical_url(response)
         db_page, created = self._get_or_create_db_page(url=url, fingerprint=fingerprint,
                                                  request_or_response=response)
         if created:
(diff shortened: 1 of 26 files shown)


sibiryakov:
@shaneaevans I really need your input here ;)

crawlfrontier/contrib/backends/sqlalchemy/__init__.py
```diff
@@ -122,7 +122,8 @@ def frontier_stop(self):
 
     def add_seeds(self, seeds):
         for seed in seeds:
-            db_page, _ = self._get_or_create_db_page(url=seed.url, fingerprint=seed.meta['fingerprint'],
+            url, fingerprint, _ = self.manager.canonicalsolver.get_canonical_url(seed)
```
shaneaevans: I see this pair of lines together all the time - `get_canonical_url` followed by `_get_or_create_db_page`. should there be a single method to do both?

crawlfrontier/contrib/canonicalsolvers/basic.py
```diff
@@ -0,0 +1,44 @@
...
+            redirect_fingerprints = obj.meta['redirect_fingerprints']
+            redirect_domains = obj.meta['redirect_domains']
+            redirect_urls.append(obj.url)
+            redirect_fingerprints.append(obj.meta['fingerprint'])
+            redirect_domains.append(canonical_domain)
+
+            canonical_url = redirect_urls[0]
+            canonical_fingerprint = redirect_fingerprints[0]
+            canonical_domain = redirect_domains[0]
+
```
shaneaevans: What is the motivation for using the original URL and fingerprint as the canonical one? the example given in the other issue was that of a URL shortener, which seems to suggest that we should be using the redirected TO url. I suppose this complicates the back end as it needs to do more mapping?

shaneaevans:
Idea looks good to me, I can see we'll need a more sophisticated version of this for our larger crawls. Needs tests.

crawlfrontier/contrib/canonicalsolvers/basic.py
sibiryakov: Right, this complicates backend. This PR is mainly about Frontera API changes, rather then actual canonical solver implementation.

sibiryakov:
Great! What do you think about the name: canonical solver? Is it ok from english prospective?

shaneaevans:
Name is ok from the English perspective.  I prefer a name like CannonicalURLMapper, which makes it clear that this may map URLs to other ones, not only perform transformations (e.g. like Scrapy https://github.com/scrapy/scrapy/blob/master/scrapy/utils/url.py#L38 )

PR #27, opened 2015-04-10
Added basic no-revisit backend

plafl:
Maybe the most basic backend you can have with some useful functionality. It could be a good starting point for new users or developers of crawlfrontier.


sibiryakov:
Hey @plafl I like the idea of having links db abstraction/interface in CF. Regarding the solution itself, sqlalchemy-based backend already supports no revisiting functionality. Please see here https://github.com/scrapinghub/crawl-frontier/blob/development/crawlfrontier/contrib/backends/sqlalchemy/__init__.py#L131. Memory backend has no such support, and I'm not sure if it's needed there. As I see memory backend, it's main purpose is to test spider-cf integration. If something doesn't work, the Memory Backend is easy option to try and compare.

sibiryakov:
PS: May be it's better to make PR's to development, until we merged it to master?

PR #34, opened 2015-04-21
Merge hcf backend rebased from master

gatufo:
hcf backend was moved to a different repo (backend was removed from frontier), but some framework changes still remained in this branch. 
We don't need this branch anymore as backend have been removed, but we still want the changes.
So, this is the PR rebased from master to discuss.

A brief summary of some relevant changes filtering hcf_backend ones:

**1. Added kwargs argument to frontier init, start and stop methods** 
https://github.com/scrapinghub/frontera/commit/f9c70f1320d83541042d9dc15c3480f04a64dde3

Frontier API methods `frontier_start` and `frontier_stop` now have kwargs arguments to allow custom parameters (like receiving scrapy spider when used with scrapy scheduler)
```python
def frontier_start(self, **kwargs):
         pass
```

**2. Disabled AUTO_START for scrapy scheduler** 
https://github.com/scrapinghub/frontera/commit/b54027c3d2946065d2e9c41eefee1a46174f5697

Just changed default value to False

**3. Scrapy scheduler automatically adds required middlewares** 
https://github.com/scrapinghub/frontera/commit/3b0b86b11e3d4e962e3a86dd0fc2ea96428b8846
https://github.com/scrapinghub/frontera/commit/6148ee76746385c391339e4b21c399c905c114cc

Scrapy scheduler automatically adds the required middlewares, so we pass from this frontier settings:
```python
SPIDER_MIDDLEWARES.update(
    {'frontera.contrib.scrapy.middlewares.schedulers.SchedulerSpiderMiddleware': 999},
)
DOWNLOADER_MIDDLEWARES.update(
    {'frontera.contrib.scrapy.middlewares.schedulers.SchedulerDownloaderMiddleware': 999}
)
SCHEDULER = 'frontera.contrib.scrapy.schedulers.frontier.FronteraScheduler'
```
to just this:
```python
SCHEDULER = 'frontera.contrib.scrapy.schedulers.frontier.FronteraScheduler'
```

@kalessin, @sibiryakov and everyone, please comment :)


frontera/contrib/scrapy/schedulers/__init__.py
```diff
@@ -0,0 +1,30 @@
+from scrapy.core.scheduler import Scheduler
+from scrapy.utils.misc import load_object
+
+
+DOWNLOADER_MIDDLEWARE = 'frontera.contrib.scrapy.middlewares.schedulers.SchedulerDownloaderMiddleware'
+SPIDER_MIDDLEWARE = 'frontera.contrib.scrapy.middlewares.schedulers.SchedulerSpiderMiddleware'
+
+
+class BaseFronteraScheduler(Scheduler):
```
sibiryakov: I like the intention, probably you wanted to simplify Frontera setup. Forced addition of downloader/spider middlewares makes the whole middleware stack behavior unpredictable and counter-intuitive. Keep in mind, that our SchedulerSpiderMiddleware, affects behavior of spider middlewares processing: it consumes all links, and bypasses only items. So potentially it could break the behavior in a unexpected way, especially for middlewares after inserted one.

frontera/core/manager.py
```diff
@@ -325,13 +336,13 @@ def get_next_requests(self, max_next_requests=0, **kwargs):
                                             (max_next_requests, self.n_requests, self.max_requests or '-')))
 
         # get next requests
-        next_requests = self.backend.get_next_requests(max_next_requests, **kwargs)
+        next_requests = self.backend.get_next_requests(max_next_requests, **kwargs) or []
```
sibiryakov: May be it makes more sense to force backends to return only lists? This code allows to return 0, None, False, etc.

frontera/core/manager.py
```diff
@@ -391,6 +402,9 @@ def request_error(self, request, error):
         #self.event_log_manager.page_crawled_error(processed_page, error)
         return processed_page
 
+    def make_request(self, url, method='GET', headers=None, cookies=None, meta=None):
```
sibiryakov: Why is that needed? Request model is accessible in backend, why not to instantiate it directly?

frontera/core/components.py
```diff
@@ -20,16 +20,20 @@ class Component(object):
...
         """
         Called when the frontier starts, see :ref:`starting/stopping the frontier <frontier-start-stop>`.
+
+        :param ** kwargs: Arbitrary number of extra keyword arguments.
         """
         pass
 
     @abstractmethod
-    def frontier_stop(self):
+    def frontier_stop(self, **kwargs):
```
sibiryakov: Do you really need arbitrary kw args for stopping? What use-cases you see?

frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -124,11 +134,13 @@ def process_exception(self, request, exception, spider):
...
         if not self.frontier.manager.auto_start:
-            self.frontier.start()
+            scrapy_kwargs = {'spider': spider}
+            self.frontier.start(**scrapy_kwargs)
 
     def close(self, reason):
         log.msg('Finishing frontier (%s)' % reason, log.INFO)
-        self.frontier.stop()
+        scrapy_kwargs = {'reason': reason}
+        self.frontier.stop(**scrapy_kwargs)
```
sibiryakov: why not ```self.frontier.stop(reason=reason)``` ?

frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -124,11 +134,13 @@ def process_exception(self, request, exception, spider):
     def open(self, spider):
         log.msg('Starting frontier', log.INFO)
         if not self.frontier.manager.auto_start:
-            self.frontier.start()
+            scrapy_kwargs = {'spider': spider}
+            self.frontier.start(**scrapy_kwargs)
```
sibiryakov: Could you explain the motivation of this? Why do you need spider reference in Frontera internals and what possible use-cases you see of it?

kalessin: I think that understanding how hcf backend works, some of the features may have sense (for example, the spider reference one). Check code:

https://github.com/scrapinghub/hcf-backend/blob/master/hcf_backend/backend.py

sibiryakov: @kalessin It's quite complex, and overall system design isn't clear just from code. I noticed there spider callbacks:
- ``get_producer_slot(request)``
- ``cf_make_request(fingerprint, qdata, request_cls)``
however, it's still not clear why are they there and how they used.

kalessin: Those callbacks are overrode by the particular spider in order to give custom instructions on how to create a request and how to define which slot to use. A default behaviour is just too much restrictive. These are tasks that in many situations only the spider knows exactly how to handle.

sibiryakov: What kind of custom instructions for creating request and defining which slot to use? Could you put several examples from spiders. I'm asking because spider reference in backend brakes encapsulation principle and it's not clear why at the moment.

kalessin: There are many situations in which we need customization at spider level:

- in the case of get_producer_slot(), this was needed mainly because we already defined standard ways to assign requests to slots along our project, which has other components (i.e. external script that scan our db and schedules all urls from it with some metadata in HCF qdata). We were using a scrapy middleware for handling HCF before frontera existed, so migration to frontera required to have this adaptation in order to define a standard way to assign urls to slots along all the project and external script be coordinated with sitemap or directory spiders without generating duplicated requests (HCF storage deduplicates inside the same frontier slot)

- in case of cf_make_requests, this is unavoidable for many operations like controlling the creation of request parameters (headers, method, request metadata, etc.) and request callback from qdata and url. This is impossible to be decided only by the backend. Each spider has its own needs.

- We also need access to spider for configuring frontier parametrs using spider arguments, because they are dynamic (i.e. we cannot set it via a hardcoded setting), most importantly, from which slot to read, as we have several instances of a spider, each one reading from a different slot.

sibiryakov: Regarding first two points, the general way to pass all the request information is to use request.meta dictionary in spider, and put all the slot and other metadata there and later access them in backend. Same for headers/method. You can even put additional fields in Request object, or customize Request/Response models in Frontera.

Regarding third one, frontier parameters. May be it's better to use Frontera settings file for this? Or there is something requiring to do this directly from spider code?

kalessin: «Regarding third one, frontier parameters. May be it's better to use Frontera»

As I said in the comment, they are dynamic and cannot be set via hardcoded settings. They are job dependent, so they are passed as spider argument. And in order to access the spider arguments, we need to access the spider.

sibiryakov: @kalessin Here is the idea how frontera options can be overwritten from command line.
``` scrapy crawl linkedin -s FRONTERA_HCF_CONSUMER_SLOT=linkedin1 ```
It sets Scrapy setting with prefix ``FRONTERA_`` which is detected by Frontera scheduler and set ``HCF_CONSUMER_SLOT`` setting in Frontera settings.
what do you think?

kalessin: As discussed, we don't support at the moment passing custom settings using dash API. Also that change is not enough because we need to visualize job settings and dash also does not support that.

However, we discussed yesterday with @shaneaevans and he agreed to use spider arguments as now. Using this has the advantage that it is already there, so we really don't need to add another way to configure jobs. Also, that does not imply to pass the spider to the backend. I suggested you either to instantiate the frontier in the open() method of the scrapy scheduler instead in the init(), or override frontier settings in open() method, thus allowing the scheduler to pass some configuration dict in the spider. He agreed that either alternative is good. And this does not violate neither encapsulation principle (spider is not passed to the backend) nor adds extra awareness of the backend to the spider (spider is used only as a via to pass arguments, nothing else)

sibiryakov: IMO, overriding frontier settings in open() method is more reasonable.

kalessin: ok, I will go with that

kalessin: @sibiryakov https://github.com/scrapinghub/frontera/commit/7131fa418cdea67880d2d9516a5a597c9dac465e

kalessin:
I think this PR should not be applied. Most of its changes are not really needed and review and choose what should go and what should not is becoming a pain. I will create a different PR with much more simple changes.

Issue #33, opened 2015-04-16
New DELAY_ON_EMPTY functionality on FronteraScheduler terminates crawl right at start


plafl:
While this is solved you can use this on your settings as a workaround:

    DELAY_ON_EMPTY=0.0

The problem is in `frontera.contrib.scrapy.schedulers.FrontieraScheduler`, method `_get_next_requests`.
If there are no pending requests and the test `self._delay_next_call < time()` fails, an empty list is returned which causes the crawl to terminate

sibiryakov:
Hey @plafl what's the status of this? Is it about backend returning zero results?

plafl:
@sibiryakov The problem is not that the backend does not return zero results. The problem is that the backend is not even called by FrontierScheduler. Because the backend is not called FrontierScheduler returns an empty list. Reverting to the old implementation of the method which always calls the backend if there are not pending requests would fix it, but then I will eliminate the functionality intended by _delay_next_call

sibiryakov:
I think we can find solution. How to reproduce that?

plafl:
1. Fetch this example:

        git clone https://github.com/scrapinghub/tutorials.git
        cd blog/hn_scrapper
    
2. Comment out line 5 of file `hn_scrapper/frontier_settings.py`:

        #DELAY_ON_EMPTY = 0.0

3. Launch spider:

        scrapy crawl HackerNews

sibiryakov:
This is a complex problem.

We faced with two problems here:
- Control of the finishing condition,
- Frontera scheduler willingness to fill the queue as fast and as full as possible, minimizing amount of calls to backend.

Here is what's happening in your spider+Scrapy+Frontera mix:
  1. Clean up of Frontera backend DB.
  1. First it asks Frontera backend if theres is something to download, gets zero results (because seeds aren't in db yet),
  1. Then it adds seeds (this is Scrapy engine behavior: first check next_request, if there is nothing then check start_urls)
  1. Exits, because of unsuccessful attempt to fill the queue in full at step 2.

By changing ```DELAY_ON_EMPTY``` to ```0.0``` you change Frontera's scheduler behavior, now it's allowed to ask for new requests as often as Scrapy needs at the price of hitting backend on every processed request (probably you don't care in this case).

You can follow and test all these observations by turning on logging in ```frontera_settings.py```:
 ```
LOGGING_EVENTS_ENABLED = True
LOGGING_MANAGER_ENABLED = True
[...]

PR #35, opened 2015-04-22
Quick fix of premature Scrapy spider closing on cold start

sibiryakov:
This isn't a complete, well thought-out solution, rather quick fix of https://github.com/scrapinghub/frontera/issues/33

In short I set default value of DELAY_ON_EMPTY to 0.0, to prevent delays on get_next_request calls and allowing it to hit DB as many times as needed, by default.
Users who make crawler focused on particular website, probably doesn't care about hitting DB too much, because DB will be small, likely. So the intention is to save their time, making Frontera work out of the box. For those who cares about DB load, this option is still available.


gatufo:
SGTM. Setting it as the default behaviour will be less confusing and in case someone needs it is also available.

plafl:
I'm happy with it.

PR #37, opened 2015-04-30
remove duplicated module file

kalessin:
This file must not be here. There is already a sqlalchemy/__init__.py file with same content


sibiryakov:
Thanks, @kalessin!

PR #38, opened 2015-05-01
some few changes needed to support backends in general

kalessin:
feature for scrapy scheduler

### DIFF
--- a/frontera/contrib/scrapy/schedulers/frontier.py
+++ b/frontera/contrib/scrapy/schedulers/frontier.py
@@ -76,10 +76,8 @@ class FronteraScheduler(Scheduler):
         self.stats_manager = StatsManager(crawler.stats)
         self._pending_requests = deque()
         self.redirect_enabled = crawler.settings.get('REDIRECT_ENABLED')
-        settings = ScrapySettingsAdapter(crawler.settings)
-        self.frontier = ScrapyFrontierManager(settings)
-        self._delay_on_empty = self.frontier.manager.settings.get('DELAY_ON_EMPTY')
         self._delay_next_call = 0.0
+        self.frontier = None
 
     @classmethod
     def from_crawler(cls, crawler):
@@ -119,6 +117,11 @@ class FronteraScheduler(Scheduler):
         self.stats_manager.add_request_error(error_code)
 
     def open(self, spider):
+        settings = ScrapySettingsAdapter(spider.crawler.settings)
+        settings.set_from_dict(getattr(spider, 'frontera_settings', {}))
+        self.frontier = ScrapyFrontierManager(settings)
+        self._delay_on_empty = self.frontier.manager.settings.get('DELAY_ON_EMPTY')
+
         self.frontier.set_spider(spider)
         log.msg('Starting frontier', log.INFO)
         if not self.frontier.manager.auto_start:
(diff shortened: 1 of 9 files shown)


frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -35,38 +36,38 @@ def __init__(self, stats, prefix=STATS_PREFIX):
...
 
     def set_pending_requests(self, pending_requests):
-        self._set_value('pending_requests_count', pending_requests)
+        self.set_value('pending_requests_count', pending_requests)
 
-    def _get_stats_name(self, variable):
-        return '%s/%s' % (self.prefix, variable)
-
-    def _inc_value(self, variable, count=1):
+    def inc_value(self, variable, count=1):
```
sibiryakov: As I see, you changed StatsManager API, and made it replaceable by outside code. For what reason one would replace default one?

frontera/utils/heap.py
```diff
@@ -65,3 +65,6 @@ def _extract_object(self):
             return wrapper.obj
         except IndexError:
             return None
+
+    def __len__(self):
```
sibiryakov: Perhaps, this isn't related to this PR?

frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -124,7 +128,8 @@ def process_exception(self, request, exception, spider):
     def open(self, spider):
         log.msg('Starting frontier', log.INFO)
         if not self.frontier.manager.auto_start:
-            self.frontier.start()
+            settings_override = getattr(spider, 'frontier_settings', {})
```
sibiryakov: I think allowing to change Frontera manager settings on the fly creates a surface for potential hardly debuggable problems. What about instantiating ``self.frontier`` in ``open()`` with already adjusted settings?

kalessin: @sibiryakov in the PR I closed you said that you preferred the way I actuall did :-/

kalessin: https://github.com/scrapinghub/frontera/pull/34#discussion_r29278778

frontera/utils/heap.py
kalessin: yes, it is. It is actually part of what we need in HCF middleware for controlling the next requests logic

frontera/contrib/scrapy/schedulers/frontier.py
kalessin: in order to allow to use stats manager from the backend for customized stats.

frontera/contrib/scrapy/schedulers/frontier.py
sibiryakov: Sorry for that. I just walked through code again, and found that instantiation in ``open()`` should be cheaper from many prospectives: there is no need to modify frontier APIs and disabling auto start. Unfortunately, that wasn't clear for me since the beginning.

sibiryakov: I also found another interesting issue. Now, we do not expect that someone will be running two or more spiders within the same Scrapy process, however Scrapy allows that http://doc.scrapy.org/en/master/topics/practices.html#running-multiple-spiders-in-the-same-process. Luckily, I don't see any problems with such a setup, but user would need to supply each spider with different storage backend settings.

Issue #30, opened 2015-04-14
CrawlSpider doesn't work with most backends


sibiryakov:
Memory-based backends are exclusion.
[CrawlSpider](https://github.com/scrapy/scrapy/blob/c9d7386a32aeb4bc7fe9654d194651eee1ede56c/scrapy/contrib/spiders/crawl.py#L30) makes use of callbacks stored in requests. These callbacks are discarded when saved into DB by Frontera. Possible solution implies pickling/unpickling callback methods to binary field in DB.

sibiryakov:
@kmike graciously told us that pickle doesn't support pickling of methods, here is how Scrapy does that https://github.com/scrapy/scrapy/blob/master/scrapy/utils/reqser.py

PR #41, opened 2015-05-14
Keep metadata when converting scrapy Response objects to Frontera Response objects

plafl:
This is necessary if, for example, some spider middleware adds some kind of relevance score to a crawled page inside the Response object.

### DIFF
--- a/frontera/contrib/scrapy/converters.py
+++ b/frontera/contrib/scrapy/converters.py
@@ -47,11 +47,13 @@ class ResponseConverter(BaseResponseConverter):
     @classmethod
     def to_frontier(cls, response):
         """response: Scrapy > Frontier"""
-        return FrontierResponse(url=response.url,
-                                status_code=response.status,
-                                headers=response.headers,
-                                body=response.body,
-                                request=response.meta['frontier_request'])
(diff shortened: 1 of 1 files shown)


sibiryakov:
Hey @plafl good start! What about callbacks and doing the same for ```Request```? See also https://github.com/scrapinghub/frontera/issues/30

plafl:
@sibiryakov Requests where already keeping their metadata. About issue #30: it does not solve it since we still need to add some mechanism inside the backends to store the metadata on disk.

sibiryakov:
If possible, could you please squash all the commits to one? It would be good for commits history. Otherwise, I see no problem with merging it.

sibiryakov:
You could also write a test: creating request, converting it to frontier and back and check if meta preserved, then creating response, converting it to frontier and check the meta contents.

plafl:
@sibiryakov Done squashing commits. Does PR #43 already test this? Anyway, I have spiders running with this right now, so it works.

sibiryakov:
It's quite likely working! I want to prevent braking it, by implementing test case. PR doesn't test this. If you want, I can implement testing of this code there, np, let me know.

PR #43, opened 2015-05-20
Persistent .meta and callbacks in Frontera backend

sibiryakov:
By carefully passing serialized callbacks from Scrapy requests to backend, persisting it, and passing back.

Solves https://github.com/scrapinghub/frontera/issues/30

BTW, I had to change the architecture of Request/Response convertors. Now each convertor should be instantiated in subclass of FrontierManagerWrapper.

### DIFF
--- a/frontera/contrib/scrapy/converters.py
+++ b/frontera/contrib/scrapy/converters.py
@@ -11,48 +11,46 @@ class RequestConverter(BaseRequestConverter):
     def __init__(self, spider):
         self.spider = spider
 
-    def to_frontier(self, request):
+    def to_frontier(self, scrapy_request):
         """request: Scrapy > Frontier"""
-        if isinstance(request.cookies, dict):
-            cookies = request.cookies
+        if isinstance(scrapy_request.cookies, dict):
+            cookies = scrapy_request.cookies
         else:
-            cookies = dict(sum([d.items() for d in request.cookies], []))
-        cb = request.callback
+            cookies = dict(sum([d.items() for d in scrapy_request.cookies], []))
+        cb = scrapy_request.callback
         if callable(cb):
             cb = _find_method(self.spider, cb)
-        eb = request.errback
+        eb = scrapy_request.errback
         if callable(eb):
             eb = _find_method(self.spider, eb)
         meta = {
             'scrapy_callback': cb,
             'scrapy_errback': eb,
+            'scrapy_meta': scrapy_request.meta,
             'origin_is_frontier': True,
         }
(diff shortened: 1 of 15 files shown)


sibiryakov:
@gatufo @kmike @plafl guys, please have a look!

plafl:
@sibiryakov Looks good to me. I suppose trying to pickle the objects is the best we can do. Have you tried running this with a CrawlSpider? I say it because not all python objects can be automatically pickled and I wonder how often this is going to happen and I don't know what is going to happen in that case, maybe raising some exception. Anyway, raising an exception is better than silently discarding meta data so let's merge it.

frontera/utils/converters.py
```diff
@@ -1,11 +1,9 @@
 class BaseRequestConverter(object):
     """Converts between frontera and XXX request objects"""
-    @classmethod
     def to_frontier(cls, request):
```
kmike: thefirst argument of regular methods should be named self, not cls

frontera/contrib/scrapy/converters.py
```diff
@@ -8,15 +8,24 @@
...
+        eb = request.errback
+        if callable(eb):
+            eb = _find_method(self.spider, eb)
         meta = {
-            'scrapy_callback': request.callback,
+            'scrapy_callback': cb,
+            'scrapy_errback': eb,
             'origin_is_frontier': True,
         }
         meta.update(request.meta or {})
```
kmike: We should put a serialized data under a reserved key in this meta. Otherwise users will be able to override internal flags like 'origin_is_frontier' or 'scrapy_callback'.

sibiryakov: ok! @kmike how do you see this?

sibiryakov:
@plafl Yes, I tried CrawlSpider, it works good!

frontera/contrib/scrapy/converters.py
kmike: something like 
```
meta = {
    'scrapy_callback': cb,
    'scrapy_errback': eb,
    'scrapy_meta': request.meta,
    'origin_is_frontier': True,
}
```

sibiryakov:
Hey @plafl, this commit https://github.com/sibiryakov/crawl-frontier/commit/d7d7dd76824342c12020845df5032dc678cf62e0 overwrites your changes in PR #41. But I think fix is preserved. Please check.

plafl:
@sibiryakov Don't worry, merge it. The changes in PR #41 are minimal, if there is any problem it will easy to fix.

PR #44, opened 2015-05-25
Updated dependencies in setup.py

sibiryakov:


setup.py
```diff
@@ -53,6 +53,9 @@
     },
     tests_require=[
         "pytest>=2.6.4",
-        "SQLAlchemy>=0.9.8",
```
gatufo: SQLAlchemy dependency it's needed to run the tests: https://travis-ci.org/scrapinghub/frontera/jobs/63970178#L229

Travis build will work as it installs dependencies from `requirements/tests.txt`: https://github.com/scrapinghub/frontera/blob/42fd790062d83d523ea81e2aae160f1b69293916/tox.ini#L14

I'm not sure about how removing it from `tests_require` will affect `setuptools`.

PR #45, opened 2015-05-26
adding property accessor for DELAY_ON_EMPTY setting

eliasdorneles:
Hey folks,

Here is a proposal to use an acessor to read the Scrapy specific DELAY_ON_EMPTY setting.

The motivation is that we want to be able to change its value in our custom backend after the frontier is started, depending on the value of other settings.

This is a small change, it won't affect other users but it will make our life much easier.

Does this look good?
Thank you!

### DIFF
--- a/frontera/contrib/scrapy/schedulers/frontier.py
+++ b/frontera/contrib/scrapy/schedulers/frontier.py
@@ -81,7 +81,6 @@ class FronteraScheduler(Scheduler):
             log.msg('FRONTERA_SETTINGS not found! Using default Frontera settings...', log.WARNING)
         self.frontier = ScrapyFrontierManager(frontier_settings)
 
-        self._delay_on_empty = self.frontier.manager.settings.get('DELAY_ON_EMPTY')
         self._delay_next_call = 0.0
 
     @classmethod
@@ -139,6 +138,10 @@ class FronteraScheduler(Scheduler):
     def has_pending_requests(self):
         return len(self) > 0
 
+    @property
+    def delay_on_empty(self):
+        return self.frontier.manager.settings.get('DELAY_ON_EMPTY')
+
     def _get_next_request(self):
         if not self.frontier.manager.finished and \
                 len(self) < self.crawler.engine.downloader.total_concurrency and \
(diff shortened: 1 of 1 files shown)


sibiryakov:
Great to see PR with small change set! Changing of settings on-the-fly is a bad idea, because it could lead to hardly debuggable problems. Please tell more about why you need it, and we could figure out a better solution.

eliasdorneles:
We want this because our custom backend decides a different value for DELAY_ON_EMPTY, depending on another setting (used to mean that it's a producer or a consumer).

Basically, we want to decide a new value for this DELAY_ON_EMPTY setting, depending on the value of a custom setting.

I'm open to alternatives, but I want to note that this code is not changing the setting on-the-fly -- it just allows someone who needs it to be able to do it.
I understand your concern about changing settings on-the-fly, I'm just asking to let me decide that in my code base. :)

sibiryakov:
Well, this modification creates surface for dirty hacks and no other profit at all.

breno:
Hi @sibiryakov! We're trying to strike a balance between practicality and purity here.

* Modification of the scheduler will be much more cumbersome
* I don't think a fork is a good option, we might have as well gone that route from the start and keep on it, and in the end SH's biggest Frontera user (DS) would keep on using a fork and not upstream. We don't want to do that, but we want to address urgent problems that we have. So again, we're talking about balance here.
* This change will only affect, as @eliasdorneles has stated, users of the DS hcf-backend, which is just us. And even then, only when we explicitly decide to do so.

bosnj:
Hi @sibiryakov!

We can make the scheduler read the settings from the spider. But if those settings need to be decided in the backend, depending on some internal logic, can those settings still be applied in the scheduler? Because if the answer is no, I don't see how any of other suggestions resolve the issue? The only way then would be to move the logic that decides those settings out of the backend.

sibiryakov:
@breno and @bosnj yesterday in private chat we agreed with @eliasdorneles that we're going to implement support of Frontera configuration via Scrapy settings module. That should solve all your problems. Elias should make a PR in the near future.

breno:
@sibiryakov thank you :)

PR #46, opened 2015-05-29
Use crawler settings as a fallback when there's no FRONTERA_SETTINGS

josericardo:
This is a follow up to https://github.com/scrapinghub/frontera/pull/45.

It enables the manager to receive the crawler settings and then instantiate
the frontera settings accordingly. I added a few tests that should make the new behavior a little clearer.

Is something along this lines acceptable? How can it be improved?


sibiryakov:
Jose, it's good that you started this! However there are some things to be done. I think, it's good to declare our goals, so you could follow my thoughts better.
Our goals here are:
* Keep Frontera independent from Scrapy,
* don't brake already built applications using Frontera and Scrapy.

Therefore, let's keep all the Scrapy code, or Scrapy-aware code under ``contrib/scrapy``.

frontera/core/manager.py
```diff
@@ -116,14 +116,14 @@ def __init__(self, request_model, response_model, backend, logger, event_log_man
             self.start()
 
     @classmethod
-    def from_settings(cls, settings=None):
+    def from_settings(cls, settings=None, crawler_settings=None):
```
sibiryakov: Here you introduce dependency on Scrapy for the whole Frontera.

frontera/settings/__init__.py
```diff
@@ -60,3 +56,30 @@ def set(self, key, value):
...
         for name, value in attributes.items():
             self.set(name, value)
+
+
+class DefaultSettings(FronteraSettings):
+    def __init__(self):
+        super(DefaultSettings, self).__init__(default_settings)
+
+
+class Settings(FronteraSettings):
```
sibiryakov: That code doesn't work when settings are retrieved by attribute.
```python
	  File "/usr/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1253, in unwindGenerator
	    return _inlineCallbacks(None, gen, Deferred())
	  File "/usr/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1107, in _inlineCallbacks
	    result = g.send(result)
	  File "/Users/sibiryakov/src/sh/scrapy/scrapy/crawler.py", line 66, in start
	    yield self.engine.open_spider(self._spider, self._start_requests())
	  File "/usr/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1253, in unwindGenerator
	    return _inlineCallbacks(None, gen, Deferred())
	--- <exception caught here> ---
	  File "/usr/local/lib/python2.7/site-packages/twisted/internet/defer.py", line 1107, in _inlineCallbacks
	    result = g.send(result)
	  File "/Users/sibiryakov/src/sh/scrapy/scrapy/core/engine.py", line 224, in open_spider
	    scheduler = self.scheduler_cls.from_crawler(self.crawler)
	  File "/Users/sibiryakov/src/sh/crawl-frontier/frontera/contrib/scrapy/schedulers/frontier.py", line 91, in from_crawler
	    return cls(crawler)
	  File "/Users/sibiryakov/src/sh/crawl-frontier/frontera/contrib/scrapy/schedulers/frontier.py", line 84, in __init__
	    crawler_settings=crawler.settings)
	  File "/Users/sibiryakov/src/sh/crawl-frontier/frontera/utils/managers.py", line 8, in __init__
	    self.manager = FrontierManager.from_settings(settings=settings, crawler_settings=crawler_settings)
	  File "/Users/sibiryakov/src/sh/crawl-frontier/frontera/core/manager.py", line 127, in from_settings
	    return FrontierManager(request_model=manager_settings.REQUEST_MODEL,
	  File "/Users/sibiryakov/src/sh/crawl-frontier/frontera/settings/__init__.py", line 28, in __getattr__
	    return self.__dict__[name]
	exceptions.KeyError: 'REQUEST_MODEL'
```

frontera/core/manager.py
josericardo: Ok @sibiryakov, thank you for the feedback. I'm working on the changes you suggested.

frontera/settings/__init__.py
josericardo: I changed __getattr__ to rely on the get method so it doesn't have to be rewritten on subclasses.

josericardo:
@sibiryakov I've updated the code and the build is green again. Please, let me know if you have any other concerns/advices.

frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -76,11 +77,14 @@ def __init__(self, crawler):
         self._pending_requests = deque()
         self.redirect_enabled = crawler.settings.get('REDIRECT_ENABLED')
 
-        frontier_settings = crawler.settings.get('FRONTERA_SETTINGS', None)
-        if not frontier_settings:
+        frontera_settings = crawler.settings.get('FRONTERA_SETTINGS', None)
```
sibiryakov: IMO, it makes sense to move these 3 lines into ``SettingsWrapper``, and access ``FRONTERA_SETTINGS`` variable from there, removing first argument to ``SettingsWrapper``.

sibiryakov:
Yeah! It looks much better now! However, it still doesn't allow to configure Frontera from Scrapy settings, but it's easy to fix ;)

``FrontierManagerWrapper.__init__`` calls ``.from_settings`` instantiating Frontera ``Settings`` and copying fields from already instantiated settings class. Settings aren't passed, I tested it.

What do you think about renaming ``SettingsWrapper`` to ``SettingsAdapter`` or ``ScrapySettingsAdapter``? It reminds me [adapter OOP pattern](http://en.wikipedia.org/wiki/Adapter_pattern) pretty much.

PS. Please don't overwrite history till we agree that it should be merged, it makes testing on my side problematic.

josericardo:
@sibiryakov I've pushed the changes you've suggested (with no history overwrites :)). I'm not sure if you agree with my approach in 214de7d. Any feedback is appreciated.

sibiryakov:
Did you tested it? At least with ``examples/scrapy_frontier`` spider or any other Scrapy spider?

josericardo:
Yes, I've played with settings running the test_scrapy_spider and things seem to be fine.

sibiryakov:
I tested it (moving settings around from Scrapy to Frontera) with my local spider, and it works :+1:

sibiryakov:
What about docs? I think we should update ``frontera-settings.rst`` and mention in ``scrapy-integration.rst`` it's possible to use Scrapy settings file and how Frontera behaves if there are settings in both files: Scrapy and original one. May be there is some other place that needs to be updated too.

frontera/contrib/scrapy/settings_adapter.py
```diff
@@ -0,0 +1,27 @@
...
+from scrapy import log
+
+
+class ScrapySettingsAdapter(BaseSettings):
+    """
+    Wrapps the frontera settings, falling back to scrapy and default settings
+    """
+    def __init__(self, crawler_settings):
+        frontera_settings = crawler_settings.get('FRONTERA_SETTINGS', None)
+        if not frontera_settings:
```
sibiryakov: May be this check is useless now? It could be quite common after we merge that PR, that there is no ``FRONTERA_SETTINGS``.

frontera/settings/__init__.py
```diff
@@ -38,7 +56,7 @@ def __setattr__(self, name, value):
             self.__dict__[name] = value
 
     def add_module(self, module):
-        if isinstance(module, Settings):
+        if isinstance(module, BaseSettings):
```
sibiryakov: That's insane! Why would we or someone else need to initialize Settings class object from Settings class object? What if we remove it?

frontera/settings/__init__.py
```diff
@@ -1,12 +1,16 @@
 import six
+import copy
```
sibiryakov: Probably, not used import.

frontera/settings/__init__.py
josericardo: Yes, since the beginning I felt this behavior weird. IMO we should be more explicit, like having a ```from_module(str)``` for those who want a new Settings instance and the rest of the code should only work with Settings objects.

josericardo: Anyway, that's how it was before the PR. Should we include it in this PR's scope?

sibiryakov:
LGTM. But, before that please squash the commits (not necessary to one, but it should look reasonable).

josericardo:
Great @sibiryakov, squashed!

sibiryakov:
Thanks, man! Well done. I hope you love :wine_glass:  in Brazil ;)

josericardo:
Thank you for your time @sibiryakov :)

Yes, we enjoy :wine_glass:, but :beers: is way more popular :D

PR #47, opened 2015-06-01
Empty scrapy_meta dict created empty by default inside frontier Reque…

plafl:
RequestConverter expects that Requests originating from Frontier contain an scrapy_meta dict inside the meta dict. The most correct way of ensuring this is creating a default empty dict at the moment of Request creation.

### DIFF
--- a/frontera/tests/test_scrapy.py
+++ b/frontera/tests/test_scrapy.py
@@ -3,6 +3,7 @@
 from frontera.contrib.scrapy.converters import RequestConverter, ResponseConverter
 from scrapy.http.request import Request as ScrapyRequest
 from scrapy.http.response import Response as ScrapyResponse
+from frontera.core.models import Request as FrontierRequest
 
 
 class TestSpider(object):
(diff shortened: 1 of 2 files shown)


sibiryakov:
What about implementing a test case for that?

plafl:
Added test case, all tests passing.

frontera/tests/test_scrapy.py
```diff
@@ -52,4 +53,8 @@ def test_request_response_converters():
     assert response_converted.meta['test_param'] == 'test_value'
     assert response_converted.url == url
     assert response_converted.status == 200
-    assert response_converted.headers['TestHeader'] == 'Test value'
\ No newline at end of file
+    assert response_converted.headers['TestHeader'] == 'Test value'
+
+    frontier_request = FrontierRequest(url)
+    request_converted = rc.from_frontier(frontier_request)
+    assert frontier_request.url == url
```
sibiryakov: You check URL here instead of .meta? If you want to check there was no exception, then this line is useless.

PR #48, opened 2015-06-10
allow to update frontera settings from spider

kalessin:
Motivation:

Scrapy spiders are very frequently configured dynamically per job, so we need a mean to configure frontera settings after spider initialization, which is where we have access to spider arguments.

Dash API does not support passing of settings per job but also, even if in future would be possible, that alone does not allow dynamic programmatic configuration inside the spider. Usually settings are low level while spider arguments are high level, so we need to compute optimal settings from the arguments passed. Of course this computation could be done from the script that schedules the spider, but it is usually spider dependent, and spider code is the best place to put spider dependent logic.

So this patch is very simple, allows lot of flexibility on spider development, does not changes settings on the fly (settings are defined before frontier instantiation) and avoids the dependence on an API to pass settings per job.

### DIFF
--- a/frontera/contrib/scrapy/schedulers/frontier.py
+++ b/frontera/contrib/scrapy/schedulers/frontier.py
@@ -76,10 +76,9 @@ class FronteraScheduler(Scheduler):
         self.stats_manager = StatsManager(crawler.stats)
         self._pending_requests = deque()
         self.redirect_enabled = crawler.settings.get('REDIRECT_ENABLED')
-        settings = ScrapySettingsAdapter(crawler.settings)
-        self.frontier = ScrapyFrontierManager(settings)
-        self._delay_on_empty = self.frontier.manager.settings.get('DELAY_ON_EMPTY')
         self._delay_next_call = 0.0
+        self.frontier = None
+        self._delay_on_empty = None
 
     @classmethod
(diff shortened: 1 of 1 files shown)


kalessin:
Will add docs once we agree on the changes

sibiryakov:
Hi Martin. Ideally, spider shouldn't be aware of environment where it's running. Therefore, I don't see profit for community (all other users, inside SH and outside), so I'm suggesting to put this patch in private repository, implementing your own Scrapy scheduler, may be subclassing from Frontera's one. Again, this is too specific to be put in public Frontera.

kalessin:
We already had this discussion. This patch does not imply spider to be aware of environment. Spider arguments are not set by the spider, but from outside it. It is the same as passing settings dynamically, but using spider arguments instead. It is the ONLY way to configure frontera per job. Spider is used only as intermediate. It is not aware of anything.

kalessin:
The concept of spider been aware of the environment is already provided by a PR you recently merged: configure frontera settings from scrapy settings. Scrapy settings can be hardcoded by spider. So the feature you don't want, not only is not the target of this patch, but it is already there.

sibiryakov:
@kalessin if you think this needs to be merged please reopen it.

PR #50, opened 2015-07-06
fix `process_spider_output` to respect scrapy generators

amleczko:
This fix the problem with converting large request generators to list.

### DIFF
--- a/frontera/contrib/scrapy/schedulers/frontier.py
+++ b/frontera/contrib/scrapy/schedulers/frontier.py
@@ -105,8 +105,7 @@ class FronteraScheduler(Scheduler):
         for element in result:
             if isinstance(element, Request):
                 links.append(element)
-            else:
-                yield element
+            yield element
         self.frontier.page_crawled(response=response,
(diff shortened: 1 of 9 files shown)


frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -107,8 +107,7 @@ def process_spider_output(self, response, result, spider):
         for element in result:
             if isinstance(element, Request):
                 links.append(element)
-            else:
-                yield element
+            yield element
```
sibiryakov: Hey @amleczko! Thanks for the PR! Unfortunately, it doesn't look like it solves the problem. IMO, the right approach would be to adopt the whole Frontera pipeline for method ``page_crawled`` in backends and manager to support generators. Can you afford this? I'm ready to guide you through if you're up for this.

breno:
@amleczko go ahead Andrew! We'd like it for you to get involved in stuff like this. cc: @sibiryakov

PR #51, opened 2015-07-13
Converter pass dont_filter

bosnj:
I believe we should pass dont_filter from Scrapy Request to Frontera Request, in order to have control over dupefilter behavior per request. Currently, Frontera middleware dupefilter checks for dont_filter, but there is no way to set it from the spider.


sibiryakov:
Hey guys!
It looks like this is too Scrapy specific. Majority of our backends are doing filtering by default and they don't check for dont_filter. How do you see they should be handling it?

bosnj:
Hi Sibiryakov,

actually this change was for the settings_override branch, because we use this value in the dupefilter implementation in this branch:

I'm not sure if it should ever be merged to master, but that would depend on the future of the dupefilter I guess. @kalessin should have a better vision about it :)

PR #57, opened 2015-08-06
fixing FRONTERA_SETTINGS variable description in docs

RajatGoyal:


sibiryakov:
Please also check L82 in these file, there should be probably module path, right?

RajatGoyal:
makes sense, changed.

sibiryakov:
Thanks!

PR #60, opened 2015-09-03
updating adding headers, method and cookies

RajatGoyal:
@sibiryakov Please Review.

### DIFF
--- a/frontera/contrib/backends/sqlalchemy/__init__.py
+++ b/frontera/contrib/backends/sqlalchemy/__init__.py
@@ -10,6 +10,7 @@ from sqlalchemy import UniqueConstraint
 
 from frontera import Backend
 from frontera.utils.misc import load_object
+from frontera.core.models import Response as frontera_response
 
 # Default settings
 DEFAULT_ENGINE = 'sqlite:///:memory:'
@@ -62,6 +63,9 @@ class Page(Base):
     state = Column(String(12))
     error = Column(String(20))
     meta = Column(PickleType())
+    headers = Column(PickleType())
+    cookies = Column(PickleType())
+    method = Column(String(6))
 
     @classmethod
     def query(cls, session):
@@ -135,7 +139,8 @@ class SQLAlchemyBackend(Backend):
         next_pages = []
         for db_page in query:
             db_page.state = Page.State.QUEUED
-            request = self.manager.request_model(url=db_page.url, meta=db_page.meta)
+            request = self.manager.request_model(url=db_page.url, meta=db_page.meta, headers=db_page.headers,
+                                                 cookies=db_page.cookies, method=db_page.method)
             next_pages.append(request)
         self.session.commit()
         return next_pages
@@ -161,9 +166,16 @@ class SQLAlchemyBackend(Backend):
         db_page.fingerprint = obj.meta['fingerprint']
         db_page.state = Page.State.NOT_CRAWLED
         db_page.url = obj.url
-        db_page.depth = 0
         db_page.created_at = datetime.datetime.utcnow()
         db_page.meta = obj.meta
+
+        if not isinstance(obj, frontera_response):
+            db_page.headers = obj.headers
+            db_page.method = obj.method
+            db_page.cookies = obj.cookies
+        else:
+            db_page.depth = 0
+
         return db_page
 
     def _get_or_create_db_page(self, obj):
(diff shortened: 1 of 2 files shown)


RajatGoyal:
@sibiryakov I have accidently committed the code from integrating scrapy settings, which I was using localy, I don't have push permissions I guess. The required change is in the backend, you can remomve the change in `frontera/contrib/scrapy/schedulers/frontier.py`

frontera/contrib/backends/sqlalchemy/__init__.py
```diff
@@ -62,6 +63,9 @@ class State:
     state = Column(String(12))
     error = Column(String(20))
     meta = Column(PickleType())
+    headers = Column(PickleType())
+    cookies = Column(PickleType())
+    method = Column(String(6))
```
sibiryakov: Why 6 symbols here? Any motivation for exactly 6? Just curious.

RajatGoyal: Methods are `GET`, `POST`, `PUT` and `DELETE`
max length is 6 for delete.

frontera/contrib/backends/sqlalchemy/__init__.py
```diff
@@ -161,9 +166,15 @@ def _create_page(self, obj):
         db_page.fingerprint = obj.meta['fingerprint']
         db_page.state = Page.State.NOT_CRAWLED
         db_page.url = obj.url
-        db_page.depth = 0
         db_page.created_at = datetime.datetime.utcnow()
         db_page.meta = obj.meta
+
+        if not isinstance(obj, frontera_response):
```
sibiryakov: Why this condition is here?

frontera/contrib/backends/sqlalchemy/__init__.py
sibiryakov: I'm not sure if we need to pass PUT/DELETE (and other like OPTIONS or CONNECT) in sqlalchemy.

sibiryakov:
@RajatGoyal Man, we need tests.

sibiryakov:
And please remove Martin's stuff from PR, by squashing and forcing push.

frontera/contrib/backends/sqlalchemy/__init__.py
RajatGoyal: I don't want to overwrite the response headers and stuff on the request headers, because maybe we would want to retry this request later, I imagine the database to be only request store and not response and response store.

sibiryakov: I think backend database is a storage of metadata records. The *record* has a wider meaning, than just requests. The important use of records is generation of requests to fetch. But, there are more uses: 
 - matching them with responses (see https://github.com/scrapinghub/frontera/issues/63#issuecomment-138244380), 
 - using them during canonical URL identification (it could require some per-host level operations),
 - additional metadata storage (for example two types of score: score for prioritization, and some content-based score, calculated during *response* parsing, used for other purposes than prioritization),
 - for statistics: what was crawled, when, what errors were encountered, etc, 
 - serve as entities to be shown in external UI.

So you can see these records as requests, but I would say this kind-of limits the whole concept.

RajatGoyal: Agreed, However, then maybe we should store both request and response information.\?

sibiryakov: I'm not sure if we need to store *response* headers and cookies. I think we should just expect that ``obj`` could be a response, and we should set headers probably from ``obj.request.headers`` in this case.

RajatGoyal: sounds good. what about depth, we would still need this check for not reseting depth.

sibiryakov: @RajatGoyal could you try to use there canonical solver? (let me know if you need help) and then if your request/response causes creation of new page you should reset depth. It's independent on the ``obj`` type.

sibiryakov:
May be it also makes sense to rename SQLiteBackend to SqlAlchemyBackend

RajatGoyal:
Yes, it does. Will change it.

sibiryakov:
Hey @RajatGoyal how is it going? I encourage you to finish it, that's a great job but needs some minor things.

RajatGoyal:
Hi @sibiryakov I have been a bit busy. I will try to finish this and others by the weekend.

sibiryakov:
@RajatGoyal any update on this?

sibiryakov:
I'll close this, and create a new ticket to continue on this.

PR #64, opened 2015-09-07
process spider level exceptions fix #63

### LINKED ISSUE #63: exception during scrapy callback maked as queued
Hi,
If there is any exception with response parsing in scrapy, the request remain matked as `QUEUED` and  no error is logged on the frontier.

RajatGoyal:
Passes spider level exceptions to the backend `process_exception` function.


sibiryakov:
This could work, but would be nice to pass:
 - Error type: where it came from, e.g. downloader or spider?
 - in case of spider, response object.

sibiryakov:
anyone here to estimate consequences of that?

PR #69, opened 2015-09-19
Feature multiple crawler support, Fix #66

### LINKED ISSUE #66: setting to switch off exception when encountering same url fingerprint
I am trying to run multiple crawlers with rdbms backend, the spiders are such that they might find the url that was visited by other spider, frontera raises as exception in this case, Is the expected behaviour?

I would want to not throw error when same url is encountered, I just want to not crawl it again, but don't need to throw exception.

Error:
```sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely) (psycopg2.IntegrityError) duplicate key value violates unique constraint "play_store_pkey"
DETAIL:  Key (fingerprint)=(f259306fa30657ab28ffa1c322d843d0cdceee41) already exists.
 [SQL: 'INSERT INTO play_store (url, fingerprint, depth, created_at, status_code, state, error, meta, headers, cookies, method, body) VALUES (%(url)s, %(fingerprint)s, %(depth)s, %(created_at)s, %(status_code)s, %(state)s, %(error)s, %(meta)s, %(headers)s, %(cookies)s, %(method)s, %(body)s)'] [parameters: {'body': None, 'cookies': <psycopg2.extensions.Binary object at 0x7f3e8a02c5a8>, 'url': 'https://play.google.com/store/apps/details?id=com.dic_o.dico_eng_fra', 'status_code': None, 'created_at': '20150909151342577212', 'error': None, 'state': 'NOT CRAWLED', 'headers': <psycopg2.extensions.Binary object at 0x7f3e89d6c238>, 'depth': 6, 'meta': <psycopg2.extensions.Binary object at 0x7f3e89d6c440>, 'fingerprint': 'f259306fa30657ab28ffa1c322d843d0cdceee41', 'method': 'GET'}]```

### PR DESCRIPTION
**Pull request by RajatGoyal**
This PR prevents integrity errors when running two or more crawlers on same database:
sqlalchemy doesn't provide any method which gives `INSERT OR IGNORE` behaviour out of the box, this approach takes a performance hit but I couldn't seem to find any other way.

If we want to handle `INSERT OR IGNORE` ignore ourself we can do:

for mysql:
 `self.session.execute(insert(self.table, values=values, prefixes=['IGNORE']))`

for sqlite:
`self.session.execute(insert(self.table, values=values, prefixes=['OR IGNORE']))`

Postgrest 9.5 has `ON CONFLICT IGNORE`. However, 9.5 is not yet common.

This approach would make the code a bit more complicated because of the raw queries.

### DIFF
--- a/frontera/contrib/backends/sqlalchemy/__init__.py
+++ b/frontera/contrib/backends/sqlalchemy/__init__.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import datetime
+from sqlalchemy import exc
 
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import create_engine
@@ -123,11 +124,11 @@ class SQLiteBackend(Backend):
 
     def add_seeds(self, seeds):
         for seed in seeds:
-            db_page, _ = self._get_or_create_db_page(seed)
+            self._create_db_page(seed, 0)
         self.session.commit()
 
     def get_next_requests(self, max_next_requests, **kwargs):
-        query = self.page_model.query(self.session)
+        query = self.page_model.query(self.session).with_lockmode('update')
         query = query.filter(self.page_model.state == Page.State.NOT_CRAWLED)
         query = self._get_order_by(query)
         if max_next_requests:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: frontera/contrib/backends/sqlalchemy/__init__.py [hunk 1]
```diff
@@ -123,11 +124,11 @@ def frontier_stop(self):
 
     def add_seeds(self, seeds):
         for seed in seeds:
-            db_page, _ = self._get_or_create_db_page(seed)
+            self._create_db_page(seed, 0)
```
sibiryakov: What if page(or document) is already exists?

frontera/contrib/backends/sqlalchemy/__init__.py
```diff
@@ -141,41 +142,38 @@ def get_next_requests(self, max_next_requests, **kwargs):
         return next_pages
 
     def page_crawled(self, response, links):
-        db_page, _ = self._get_or_create_db_page(response)
```
sibiryakov: See https://github.com/scrapinghub/frontera/pull/70#issuecomment-141939767, unfortunately without solving canonical urls that code will end up loosing all redirected urls.

frontera/contrib/backends/sqlalchemy/__init__.py
```diff
@@ -123,11 +124,11 @@ def frontier_stop(self):
 
     def add_seeds(self, seeds):
         for seed in seeds:
-            db_page, _ = self._get_or_create_db_page(seed)
+            self._create_db_page(seed, 0)
         self.session.commit()
 
     def get_next_requests(self, max_next_requests, **kwargs):
-        query = self.page_model.query(self.session)
+        query = self.page_model.query(self.session).with_lockmode('update')
```
sibiryakov: That should cause slowdown in general single process state.Would be nice to hide it with option, probably.

sibiryakov:
This use case is handled in Frontera 0.4 different way: it is recommended to use ``distributed spiders`` run mode, with a single worker communicating with DB and spiders communicating using ZeroMQ. So no heavy services needed, and the whole solution can easily fit within single machine.

sibiryakov:
Please reopen, if you it still makes sense.

PR #70, opened 2015-09-19
Marking QUEUED as NOT CRAWLED on closing

RajatGoyal:
Mark `QUEUED` as not `NOT CRAWLED`.

### DIFF
--- a/frontera/contrib/backends/sqlalchemy/__init__.py
+++ b/frontera/contrib/backends/sqlalchemy/__init__.py
@@ -118,6 +118,10 @@ class SQLiteBackend(Backend):
         pass
 
     def frontier_stop(self):
+        query = self.page_model.query(self.session).filter(self.page_model.state == Page.State.QUEUED)
+        for db_page in query:
+            db_page.state = Page.State.NOT_CRAWLED
+        self.session.commit()
         self.session.close()
         self.engine.dispose()
 


sibiryakov:
Only one thing I'm concerned is: what if we can't match initial URL with response? Say Frontera creates request with URL A, but it causes a redirect and returns as URL B. Or user has a custom buggy fingerprint function, which sometimes generates different fingerprint for the same URLs. This way the number of documents with ``QUEUED`` state will grow eventually, along with creation of new URLs with ``CRAWLED`` state (without passing common flow: NOT_CRAWLED->QUEUED->CRAWLED/ERROR).

Would be nice probably to collect stats about unmatched documents and write warnings to log about that (that should be easy to disable in settings).

sibiryakov:
Based on such stats we could write tests also ;)

sibiryakov:
@RajatGoyal So, Rajat I'm not sure if we have to merge it. This will also hide the fact that some urls weren't matched during fetching. Therefore, making debugging even harder.

sibiryakov:
@RajatGoyal in Frontera 0.4 we had a redesign of backends. So ``QUEUED`` state isn't used anymore, instead these states are stored in separate table/model ``States``. The behavior needed  can be achieved by customizing States class. So, I propose to close it.

PR #72, opened 2015-09-29
Import tldextract lazily

eliasdorneles:
This makes tldextract optional for code that doesn't use the standard URL parsing with `parse_domain_from_url(url)`.

### DIFF
--- a/docs/source/topics/frontera-settings.rst
+++ b/docs/source/topics/frontera-settings.rst
@@ -225,6 +225,21 @@ Default: ``frontera.utils.fingerprint.sha1``
 The function used to calculate the ``domain`` fingerprint.
 
 
+.. setting:: TLDEXTRACT_DOMAIN_INFO
+
+TLDEXTRACT_DOMAIN_INFO
+----------------------
+
+Default: ``False``
+
+If set to ``True``, will use `tldextract`_ to attach extra domain information
+(second-level, top-level and subdomain) to meta field (see :ref:`frontier-objects-additional-data`).
+
+
+.. _tldextract: https://pypi.python.org/pypi/tldextract
+
+
+
 Default settings
 ================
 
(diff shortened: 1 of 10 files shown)


sibiryakov:
@eliasdorneles it looks ok, let's remove the dependencies from ``requirements.txt`` and ``setup.py``. 

I would appreciate if you could implement the option in ``DomainMiddleware`` allowing to switch between native and ``tldextract`` versions of ``parse_domain_info``. That would be beneficial for the project.

sibiryakov:
Actually you could move ``tldextract`` to ``extras`` section in setup.py.

eliasdorneles:
Nice!
Yeah, I noticed that the domain middleware is actually using the version with urlparse (`parse_domain_from_url_fast`), so tldextract is optional by default.

I'll work on this later, since changes in `DomainMiddleware` will need tests.
Btw, any suggestion on how to implement the configurability? Should I add a new setting?
There is a `DOMAIN_FINGERPRINT_FUNCTION`, I suppose I could add a `DOMAIN_URL_PARSING_FUNCTION`.

sibiryakov:
I think customizing domain extraction function is a bit too much. Let's just make a turn on/off switch for extended domain info provided by ``tldextract``, what do you think?

eliasdorneles:
Sounds good to me. Hard to come up with a good naming, tho. :)
The extra details are second-level domain, top-level domain and subdomain.

`EXTRACT_EXTRA_DOMAIN_INFO` with a default to False?

sibiryakov:
What about ``EXTENDED_DOMAIN_INFO`` or ``TLDEXTRACT_DOMAIN_INFO``. Of course we should mention these details in docs.

eliasdorneles:
Right.
Okay, my vote goes to `TLDEXTRACT_DOMAIN_INFO`. :+1:

eliasdorneles:
@sibiryakov alright, I implemented the optional tldextract, made optional in setup.py, and added tests and docs.
Does this look good?

PR #80, opened 2015-11-23
Add body and dont_filter to Frontera request

vincentlaucy:
Given frontera support POST/PUT, it make sense to have the body in request. 
Also added `dont_filter` to reuse the convention that such request should not be filtered by backend (even if it is duplicated)

What is the major idea of creating another Request object specialized for Frontera?
here we assumed encoding done by scrapy, which may need further fine tune.

### DIFF
--- a/frontera/contrib/scrapy/converters.py
+++ b/frontera/contrib/scrapy/converters.py
@@ -41,8 +41,10 @@ class RequestConverter(BaseRequestConverter):
         return FrontierRequest(url=scrapy_request.url,
                                method=scrapy_request.method,
                                headers=scrapy_request.headers,
+                               body=scrapy_request.body,
                                cookies=cookies,
-                               meta=meta)
+                               meta=meta,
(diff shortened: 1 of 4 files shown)


sibiryakov:
Hi @vincentlaucy! Thanks for the PR. Yes body in request absolutely makes sense. It would be nice to support this idea in SQL backend also, since it's connected with Frontera primary use-cases.
If you don't mind please add a test case for body and documentation.

The major idea behind specialized ``Request`` object is using Frontera with other frameworks. It's possible to use Frontera with ``requests`` library or even with fetcher implemented in C/Java connecting it by means of Kafka. Next, this is the reason why ``dont_filter`` field doesn't makes much sense here, since this concept is tightly connected to Scrapy scheduler/middlewares implementation. IMO, storing ``dont_filter`` in Frontera's meta would be valuable.

sibiryakov:
Closing, because body is already in Request.

PR #84, opened 2015-11-30
Do not enforce AWS keys in settings for the S3SeedLoader

chris-zen:
`S3SeedLoader` enforces to define the AWS keys through the settings, but `boto` is able to detect them from the default AWS settings.

Not having to specify the AWS keys is very convenient when using the AWS IAM Temporary Security Credentials through the AWS Security Token Service, where AWS keys are generated automatically and for a short period of time. 

More info here: http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html

### DIFF
--- a/frontera/contrib/scrapy/middlewares/seeds/s3.py
+++ b/frontera/contrib/scrapy/middlewares/seeds/s3.py
@@ -15,7 +15,7 @@ class S3SeedLoader(FileSeedLoader):
         self.bucket_keys_prefix = u.path.lstrip('/')
         self.s3_aws_access_key = settings.get('SEEDS_AWS_ACCESS_KEY')
         self.s3_aws_secret_key = settings.get('SEEDS_AWS_SECRET_ACCESS_KEY')
-        
+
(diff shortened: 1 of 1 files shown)


sibiryakov:
Hey! Thanks for PR!
Is it friendly to production deployment? I mean configuring it via ``boto``.
May be it's worth to remove these settings at all?

chris-zen:
Hi @sibiryakov, you are welcome.

As far as I know any medium/big organization will avoid working with AWS keys directly, and will use temporary tokens instead. I would leave those setting variables but document that they are optional because boto take them from the predefined $HOME/.aws/credentials

sibiryakov:
Here is description of how boto configuration works https://boto.readthedocs.org/en/latest/boto_config_tut.html

sibiryakov:
Ok, ready to merge? Is docs needs to be updated?

sibiryakov:
We don't have a docs for S3SeedLoader, unfortunately. So, merging as is. Thank you :beer:

PR #91, opened 2016-01-06
Fix a minor bug so the colors will work.

lljrsr:
The variable used in ColorFormatter is called fmt not format.

### DIFF
--- a/frontera/logger/formatters/color.py
+++ b/frontera/logger/formatters/color.py
@@ -10,7 +10,7 @@ from colorlog import ColoredFormatter
 class ColorFormatter(ColoredFormatter):
 
     def __init__(self, format, log_colors, log_color_field, datefmt=None, reset=True, style='%'):
-        super(ColorFormatter, self).__init__(format=format, datefmt=datefmt, log_colors=log_colors,
+        super(ColorFormatter, self).__init__(fmt=format, datefmt=datefmt, log_colors=log_colors,
                                              reset=reset, style=style)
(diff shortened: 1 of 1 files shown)


sibiryakov:
Hey @lljrsr! I see you post a lot of updates! Thank you very much for this. I'm on holidays (in Russia we have one week since NY), so I'll check all that at Monday.

lljrsr:
Don't worry :) . Enjoy your holiday.

frontera/logger/formatters/color.py
```diff
@@ -10,7 +10,7 @@
 class ColorFormatter(ColoredFormatter):
 
     def __init__(self, format, log_colors, log_color_field, datefmt=None, reset=True, style='%'):
-        super(ColorFormatter, self).__init__(format=format, datefmt=datefmt, log_colors=log_colors,
+        super(ColorFormatter, self).__init__(fmt=format, datefmt=datefmt, log_colors=log_colors,
```
sibiryakov: It looks like guys changed the method signature. See https://github.com/borntyping/python-colorlog/blob/v2.6.0/colorlog/colorlog.py#L67
What if we use positional arguments for `format` instead of key-valued? Will it bring us more resistance to signature changes?

lljrsr: I do not think that this is a good idea. With a positional argument we might not even realize that the method changed and strange things could happen without us knowing why. But then again it's a matter of taste.

Issue #87, opened 2015-12-21
Redirect loop when using distributed-frontera


lljrsr:
I am using the development version of distributed-frontera, frontera and scrapy for crawling. After a while my spider keeps getting stuck in a redirect loop. Restarting the spider helps, but after a while this happens:
```
2015-12-21 17:23:22 [scrapy] INFO: Crawled 10354 pages (at 0 pages/min), scraped 0 items (at 0 items/min)
2015-12-21 17:23:22 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
2015-12-21 17:23:22 [scrapy] DEBUG: Redirecting (302) to <GET http://www.toraath.com/index.php?lid=35&name=Downloads&req=getit> from <GET http://www.toraath.com/index.php?lid=35&name=Downloads&req=getit>
2015-12-21 17:23:22 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
2015-12-21 17:23:23 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
2015-12-21 17:23:24 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
2015-12-21 17:23:25 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
2015-12-21 17:23:25 [scrapy] DEBUG: Redirecting (302) to <GET http://www.toraath.com/index.php?lid=35&name=Downloads&req=getit> from <GET http://www.toraath.com/index.php?lid=35&name=Downloads&req=getit>
2015-12-21 17:23:25 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
2015-12-21 17:23:26 [scrapy] DEBUG: Redirecting (302) to <GET http://www.reg.ru/domain/new-gtlds> from <GET http://www.reg.ru/domain/new-gtlds>
[...]

PR #94, opened 2016-01-14
Fix DW pushing same request multiple times.

lljrsr:
DW was pushing multiple requests at the same time. This is related to issue #87 . However I did not test if redirects are working yet.

### DIFF
--- a/frontera/contrib/backends/hbase.py
+++ b/frontera/contrib/backends/hbase.py
@@ -223,10 +223,12 @@ class HBaseQueue(Queue):
 
         results = []
         trash_can = set()
+
         for _, fprints in queue.iteritems():
             for fprint in fprints:
                 for rk, _ in meta_map[fprint]:
-                    trash_can.add(rk)
+                    if rk in trash_can:
+                        continue
                     for rk_fprint in fprint_map[rk]:
                         _, item = meta_map[rk_fprint][0]
                         _, _, url, score = item
@@ -234,6 +236,7 @@ class HBaseQueue(Queue):
                             'fingerprint': hexlify(rk_fprint),
                             'score': score,
                         }))
+                    trash_can.add(rk)
 
         with table.batch(transaction=True) as b:
             for rk in trash_can:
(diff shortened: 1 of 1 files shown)


sibiryakov:
I would start with investigating how these URLs passed the SW state cache. Scheduling of multiple URLs shouldn't happen.

lljrsr:
I do not think that SW are pushing multiple URLs. I think that this is a bug in HBase Backend where each row key is added multiple times to the new batch.

frontera/contrib/backends/hbase.py
```diff
@@ -223,17 +223,19 @@ def get_next_requests(self, max_n_requests, partition_id, **kwargs):
...
                 for rk, _ in meta_map[fprint]:
-                    trash_can.add(rk)
-                    for rk_fprint in fprint_map[rk]:
-                        _, item = meta_map[rk_fprint][0]
-                        _, _, url, score = item
-                        results.append(Request(url, meta={
-                            'fingerprint': hexlify(rk_fprint),
-                            'score': score,
-                        }))
+                    if rk not in trash_can:
```
sibiryakov: What if we rewrite it:
```python
if rk in trash_can:
    continue
```
It will prevent growing of indent.

sibiryakov:
You were right actually. It generates plenty of duplicates, perhaps I left this error after refactoring to metadata/queue/states model.

sibiryakov:
All items stored at some RK, were collected to ``results`` list len(rk) times...

lljrsr:
Awesome :) . So let's merge this with master. I was a bit unsure about renaming items to requests, since there are so many different mentionings of items in the whole project. Maybe that is something you could do in a different branch to differentiate between requests and items.

sibiryakov:
Your renaming was a right thing here. Because *items* related to Items Pipeline in Scrapy, which is absolutely different thing.

sibiryakov:
Could you please rebase your branch, merging everything to one commit and putting clear commit message? It could be done with
```bash
$git rebase -i HEAD~4 (you have 4 commits in your branch)
$git log (check if it looks ok, check the code)
$git push -f
```

lljrsr:
It looks like everything worked fine :) .

sibiryakov:
Great :beers: !

PR #95, opened 2016-01-15
Change SPIDER_FEED_PARTITIONS doc value

### LINKED ISSUE #89: Default settings and documentation inconsistent
Hi there.
In the [default settings](https://github.com/scrapinghub/frontera/blob/5dd126294cd4c21c77500fc618d603c65e229fa0/frontera/settings/default_settings.py#L35) you set
```
SPIDER_FEED_PARTITIONS = 1
```
In the [docs](https://github.com/scrapinghub/frontera/blob/5dd126294cd4c21c77500fc618d603c65e229fa0/docs/source/topics/frontera-settings.rst#spider_feed_partitions) it is written that the default is supposed to be 2.
I also think that the docs about SPIDER_FEED_PARTIONS in the [run modes section](https://github.com/scrapinghub/frontera/blob/5dd126294cd4c21c77500fc618d603c65e229fa0/docs/source/topics/run-modes.rst#distributed-spiders) are a bit confusing. I think that setting `SPIDER_FEED_PARTIONS = 2` will result in two partitions with the numbers 0 and 1. According to this docs page it should be 0, 1 and 2.

lljrsr:
Fixes #89.

### DIFF
--- a/docs/source/topics/frontera-settings.rst
+++ b/docs/source/topics/frontera-settings.rst
@@ -259,7 +259,7 @@ each strategy worker assigned to it's own partition.
 SPIDER_FEED_PARTITIONS
 ----------------------
 
-Default: ``2``
+Default: ``1``
 
 Number of :term:`spider feed` partitions. This directly affects number of spider processes running. Every spider is
 assigned to it's own partition.


sibiryakov:
thanks!

PR #108, opened 2016-02-04
Resolve hostname to ip.

### LINKED ISSUE #99: ZMQ_HOSTNAME cannot be a host name when connecting to broker
When using a host name to connect via ZMQ an error is thrown because the socket tries to connect to the host name, but needs an IP address instead.

lljrsr:
Fix #99 in a hopefully nice way :zap:

### DIFF
--- a/docs/source/topics/frontera-settings.rst
+++ b/docs/source/topics/frontera-settings.rst
@@ -532,14 +532,15 @@ ZeroMQ message bus settings
 
 The message bus class is ``distributed_frontera.messagebus.zeromq.MessageBus``
 
-.. setting:: ZMQ_HOSTNAME
+.. setting:: ZMQ_ADDRESS
 
-ZMQ_HOSTNAME
+ZMQ_ADDRESS
 ------------
 
 Default: ``127.0.0.1``
 
-Hostname, where ZeroMQ socket should bind or connect.
+Defines where the ZeroMQ socket should bind or connect. Can be a hostname or an IP
+address.
 
 .. setting:: ZMQ_BASE_PORT
 
--- a/docs/source/topics/frontera-settings.rst
+++ b/docs/source/topics/frontera-settings.rst
@@ -540,7 +540,8 @@ ZMQ_ADDRESS
 Default: ``127.0.0.1``
 
 Defines where the ZeroMQ socket should bind or connect. Can be a hostname or an IP
-address.
+address. Right now ZMQ has only been properly tested with IPv4. Proper IPv6
+support will be added in the near future.
 
 .. setting:: ZMQ_BASE_PORT
 
--- a/frontera/contrib/messagebus/zeromq/broker.py
+++ b/frontera/contrib/messagebus/zeromq/broker.py
@@ -26,7 +26,7 @@ class Server(object):
     db_in = None
     db_out = None
 
-    def __init__(self, hostname, base_port):
+    def __init__(self, address, base_port):
         self.ctx = zmq.Context()
         self.loop = IOLoop.instance()
         self.stats = {
@@ -39,7 +39,10 @@ class Server(object):
             'sw_out_recvd': 0
         }
 
-        socket_config = SocketConfig(hostname, base_port)
+        socket_config = SocketConfig(address, base_port)
+
+        if socket_config.is_ipv6:
+            self.ctx.setsockopt(zmq.IPV6, True)
 
         spiders_in_s = self.ctx.socket(zmq.XPUB)
         spiders_out_s = self.ctx.socket(zmq.XSUB)
@@ -69,8 +72,12 @@ class Server(object):
         self.sw_in.on_recv(self.handle_sw_in_recv)
         self.db_in.on_recv(self.handle_db_in_recv)
         self.spiders_in.on_recv(self.handle_spiders_in_recv)
-        logging.basicConfig(format="%(asctime)s %(message)s", datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO)
-        self.logger = logging.getLogger("distributed_frontera.messagebus.zeromq.broker.Server")
+        logging.basicConfig(format="%(asctime)s %(message)s",
+                            datefmt="%Y-%m-%d %H:%M:%S", level=logging.INFO)
+        self.logger = logging.getLogger("distributed_frontera.messagebus"
+                                        ".zeromq.broker.Server")
+        self.logger.info("Using socket: {}:{}".format(socket_config.ip_addr,
+                                                      socket_config.base_port))
 
     def start(self):
         self.logger.info("Distributed Frontera ZeroMQ broker is started.")
@@ -138,21 +145,30 @@ class Server(object):
 
 
 def main():
+    """
+    Parse arguments, set configuration values, then start the broker
+    """
     parser = ArgumentParser(description="Crawl frontier worker.")
-    parser.add_argument('--config', type=str,
-                        help='Settings module name, should be accessible by import.')
-    parser.add_argument('--hostname', type=str,
-                        help='Hostname or IP address to bind. Default is 127.0.0.1')
-    parser.add_argument('--log-level', '-L', type=str, default='INFO',
-                        help='Log level, for ex. DEBUG, INFO, WARN, ERROR, FATAL. Default is INFO.')
-    parser.add_argument('--port', type=int,
-                        help='Base port number, server will bind to 6 ports starting from base. Default is 5550')
+    parser.add_argument(
+        '--config', type=str,
+        help='Settings module name, should be accessible by import.')
+    parser.add_argument(
+        '--address', type=str,
+        help='Hostname, IP address or Wildcard * to bind. Default is 127.0.0.1')
+    parser.add_argument(
+        '--log-level', '-L', type=str, default='INFO',
+        help='Log level, for ex. DEBUG, INFO, WARN, ERROR, FATAL. Default is'
+        ' INFO.')
+    parser.add_argument(
+        '--port', type=int,
+        help='Base port number, server will bind to 6 ports starting from base'
+        '. Default is 5550')
     args = parser.parse_args()
 
     settings = Settings(module=args.config)
-    hostname = args.hostname if args.hostname else settings.get("ZMQ_HOSTNAME")
+    address = args.address if args.address else settings.get("ZMQ_ADDRESS")
     port = args.port if args.port else settings.get("ZMQ_BASE_PORT")
-    server = Server(hostname, port)
+    server = Server(address, port)
     server.logger.setLevel(args.log_level)
     server.start()
 
--- a/frontera/contrib/messagebus/zeromq/broker.py
+++ b/frontera/contrib/messagebus/zeromq/broker.py
@@ -154,7 +154,8 @@ def main():
         help='Settings module name, should be accessible by import.')
     parser.add_argument(
         '--address', type=str,
-        help='Hostname, IP address or Wildcard * to bind. Default is 127.0.0.1')
+        help='Hostname, IP address or Wildcard * to bind. Default is 127.0.0.1'
+        '. When binding to wildcard it defaults to IPv4.')
     parser.add_argument(
         '--log-level', '-L', type=str, default='INFO',
         help='Log level, for ex. DEBUG, INFO, WARN, ERROR, FATAL. Default is'
--- a/frontera/contrib/messagebus/zeromq/socket_config.py
+++ b/frontera/contrib/messagebus/zeromq/socket_config.py
@@ -27,7 +27,7 @@ class SocketConfig(object):
 
     def spiders_in(self):
         """
-        TCP socket for ingoing spider messages
+        TCP socket for incoming spider messages
         """
         return 'tcp://%s:%d' % (self.ip_addr, self.base_port)
 
@@ -39,7 +39,7 @@ class SocketConfig(object):
 
     def sw_in(self):
         """
-        TCP socket for ingoing SW messages
+        TCP socket for incoming SW messages
         """
         return 'tcp://%s:%d' % (self.ip_addr, self.base_port + 2)
 
@@ -51,7 +51,7 @@ class SocketConfig(object):
 
     def db_in(self):
         """
-        TCP socket for ingoing messages
+        TCP socket for incoming messages
         """
         return 'tcp://%s:%d' % (self.ip_addr, self.base_port + 4)
 
--- a/frontera/contrib/messagebus/zeromq/socket_config.py
+++ b/frontera/contrib/messagebus/zeromq/socket_config.py
@@ -1,29 +1,62 @@
 # -*- coding: utf-8 -*-
+"""
+Contains the SocketConfig class
+"""
+from socket import getaddrinfo, gaierror
 
 
 class SocketConfig(object):
-
-    hostname = None
-    base_port = None
-
-    def __init__(self, hostname, base_port):
-        self.hostname = hostname
-        self.base_port = base_port
+    """
+    Converts address to IPv4 or IPv6 or * and returns the necessary socket
+    addresses.
+    NOTE: When using * it defaults to IPv4
+    """
+    def __init__(self, address, base_port):
+        if address == '*':
+            self.ip_addr = '*'
+            self.base_port = base_port
+            self.is_ipv6 = False
+        else:
+            try:
+                addr_tuple = getaddrinfo(address, base_port)[0][4]
+            except gaierror:
+                raise gaierror("Hostname '%s' could not be resolved" % address)
+            self.ip_addr = addr_tuple[0]
+            self.base_port = addr_tuple[1]
+            self.is_ipv6 = True if len(addr_tuple) == 4 else False
 
     def spiders_in(self):
-        return 'tcp://%s:%d' % (self.hostname, self.base_port)
+        """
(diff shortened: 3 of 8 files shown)


frontera/contrib/messagebus/zeromq/socket_config.py
```diff
@@ -1,29 +1,33 @@
 # -*- coding: utf-8 -*-
+import socket
 
 
 class SocketConfig(object):
-
-    hostname = None
-    base_port = None
+    __ip_addr = None
```
sibiryakov: Are you sure we need a [name mangling](http://stackoverflow.com/questions/1301346/the-meaning-of-a-single-and-a-double-underscore-before-an-object-name-in-python) here?  Maybe one underscore, would be the right thing to do?

lljrsr: It is not needed here, but I think it is a good convention to program as restrictive as possible. That way strange errors like name overlaps can be prohibited.
However since this is your code and you do not seem to like it, I will change it.
Still.. do you have any input on why you prefer `_` instead of `__`

frontera/contrib/messagebus/zeromq/socket_config.py
```diff
@@ -1,29 +1,33 @@
...
+    __base_port = None
 
     def __init__(self, hostname, base_port):
-        self.hostname = hostname
-        self.base_port = base_port
+        try:
+            self.__ip_addr = socket.gethostbyname(hostname)
```
sibiryakov: This wouldn't work with IPv6.

sibiryakov:
I just realized that this class is broken by design: because it tries to fit binding and connecting to port in one class, which isn't possible in general.
For binding I think we have to provide:
 - wild-card * (listen on all interfaces),
 - specifying interface by IP (both IPv4 or IPv6),
 - interface name (optional, IMO).

When connecting:
- DNS name,
- IPv6 or IPv4 of destination address.

I would go with different settings keys for where to bind and where to connect, and leave the one key for specifying port range interval start. Perhaps, ``SocketConfig`` class could provide a way for broker, to figure out where/how to bind, and current interface can be used by clients without change.

Sorry man, it looks more complex when you expect probably. It would be great to cover it with test case also.

sibiryakov:
This is actually pretty high priority, in fact, current implementation allows to operate localhost-only :no_good:

lljrsr:
I use it with IPv4 addresses. They seems to be working.

frontera/contrib/messagebus/zeromq/socket_config.py
sibiryakov: @lljrsr this code isn't mine. I'm just looking after it, because I can't just leave it for free modification to everyone.
I think these private methods can be used in possible descendants.

sibiryakov:
With different config files, yes.

lljrsr:
What do you mean by different config files? I have one settings file for all workers (and broker) and one settings file for all spiders. The both have the same IP for `ZMQ_HOSTNAME`.
I also have a `scrapy.cfg` file, but using that is not supported by frontera by default.

lljrsr:
I do not think that frontera users need all of those options when connecting/binding to ZMQ. Giving them the option to either use an IP address or a hostname should be sufficient. So the setting `ZMQ_HOST` should be sufficient in my opinion.
I have proposed a fix which at least makes binding to IPv6 work (IPv4 binding and connecting is still working). However I do not seem to get connecting to IPv6 to work. Since I am not that familiar with ZMQ maybe you could have a look?

frontera/contrib/messagebus/zeromq/socket_config.py
lljrsr: Mh.. you could them to `_` as soon as you create those descendants.

sibiryakov:
> The both have the same IP for ZMQ_HOSTNAME.
That looks like a localhost emulation. I meant a different scenario: broker, worker and spiders are all situated on separate machines within the same subnetwork. So, you could get this to work only with different ZMQ_HOSTNAME options for every instance.

frontera/contrib/messagebus/zeromq/broker.py
```diff
@@ -48,6 +48,14 @@ def __init__(self, hostname, base_port):
         db_in_s = self.ctx.socket(zmq.XPUB)
         db_out_s = self.ctx.socket(zmq.XSUB)
 
+        if socket_config.is_ipv6:
+            spiders_in_s.ipv6 = True
```
sibiryakov: I can't find this property in documentation: https://pyzmq.readthedocs.org/en/latest/api/zmq.html#socket. Am I look in the wrong place?

sibiryakov:
@lljrsr it would be nice to test it.

frontera/contrib/messagebus/zeromq/broker.py
lljrsr: I also could not find it in the docs. I found it [here](https://github.com/zeromq/libzmq/issues/853).

lljrsr:
> So, you could get this to work only with different ZMQ_HOSTNAME options for every instance.

I am running frontera on different virtual machines in my network. Each of those machines has a different IP address. I only run one zmq broker on one machine. The IP address of this machine is what I used for `ZMQ_HOSTNAME` and is the same for every machine in my network.

> @lljrsr it would be nice to test it. 

First I want to confirm that you think this way of solving it may be a good idea because testing it might be time consuming.

lljrsr:
I now use the correct zmq socket option for IPv6. I also found out that I forgot the same in the Producer and Consumer class. I tried it with `ip6-localhost` as `ZMQ_HOSTNAME`. Everything works fine now.
What do you think?

lljrsr:
Mh. The tests do not seem to pass. It may be that old ZMQ versions will not accept `tcp://[%s]:%d' % (self._ip_addr, self._base_port + 2)`
In the logs it says: 'Warning: Couldn't find an acceptable libzmq on the system.'. Do you know which ZMQ version you are testing against? In my ZMQ it works fine.

lljrsr:
Now it is working :) .

frontera/contrib/messagebus/zeromq/socket_config.py
```diff
@@ -1,29 +1,33 @@
...
-        self.base_port = base_port
+        try:
+            addr_tuple = getaddrinfo(hostname, base_port)[0][4]
+            self._ip_addr = addr_tuple[0]
+            self._base_port = addr_tuple[1]
+        except gaierror:
+            raise gaierror("Hostname '%s' could not be resolved" % hostname)
+
+    def is_ipv6(self):
+        return self._ip_addr.rfind(':')
```
sibiryakov: I think the right way is to avoid detecting IPv6 address that way. ``getaddrinfo`` returns a tuple, from which you could identify if hostname is an IPv4 or 6. Also, it handles all parsing errors, if there is wrong format (or wrong numbers) of ip address.

sibiryakov: https://docs.python.org/2/library/socket.html?highlight=getaddrinfo#socket.getaddrinfo
 >(a (address, port) 2-tuple for AF_INET, a (address, port, flow info, scope id) 4-tuple for AF_INET6)

sibiryakov:
@lljrsr you wrote
>  The IP address of this machine is what I used for ZMQ_HOSTNAME and is the same for every machine in my network.
IP Address is the same for all machines in your network? What that means? I'm not sure if machines could reside in the same subnet with equal IP addresses without conflicts.

sibiryakov:
@lljrsr and yes, overall approach looks fine!
I would love also to have a wild-card option, when running broker.

lljrsr:
>IP Address is the same for all machines in your network? What that means? I'm not sure if machines could reside in the same subnet with equal IP addresses without conflicts.

I have one machine which runs one ZMQ broker. I am referring to this machine. The other machines have other IP addresses.

sibiryakov:
Let me check it again.

sibiryakov:
Let's add a wild-card * possibility for binding to all available interfaces. I want this mainly because, we're building open source software, and it's hard to predict how and in which conditions it will be used, so that's why I think providing max options is a good idea.

sibiryakov:
Otherwise looks good. Could you also write a test case and docs update for options?

lljrsr:
Is the wild-card not working? As I said.. on my standalone setup it already works (have not tested it on my distributed setup though).

sibiryakov:
```python
>>> from socket import getaddrinfo
>>> getaddrinfo('*', 5555)
  File "<stdin>", line 1, in <module>
socket.gaierror: [Errno 8] nodename nor servname provided, or not known
```

lljrsr:
Mhh...
```
Python 2.7.10 (default, Oct 14 2015, 16:09:02) 
[GCC 5.2.1 20151010] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from socket import getaddrinfo
>>> getaddrinfo('*', 5555)
[(10, 1, 6, '', ('::1', 5555, 0, 0)), (10, 2, 17, '', ('::1', 5555, 0, 0)), (10, 3, 0, '', ('::1', 5555, 0, 0)), (2, 1, 6, '', ('127.0.0.1', 5555)), (2, 2, 17, '', ('127.0.0.1', 5555)), (2, 3, 0, '', ('127.0.0.1', 5555))]
```

sibiryakov:
```bash
$ python
Python 2.7.10 (default, Sep 23 2015, 04:34:14)
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
```

So, it's platform dependent. Ok, than it should work. What about test and docs?

lljrsr:
I will do that :) .

lljrsr:
The tests should now work.. soo what do you think about my tests? Do you have a better approach on how to test if zmq is working properly?

docs/source/topics/frontera-settings.rst
```diff
@@ -532,14 +532,15 @@ ZeroMQ message bus settings
...
 
-.. setting:: ZMQ_HOSTNAME
+.. setting:: ZMQ_ADDRESS
 
-ZMQ_HOSTNAME
+ZMQ_ADDRESS
 ------------
 
-Default: ``127.0.0.1``
+Default: ``localhost``
```
sibiryakov: Defaults to IP is more reliable from cross-platform reliability prospective.

.travis.yml
```diff
@@ -22,3 +22,6 @@ deploy:
     branch: master
     tags: true
     condition: $TOXENV == py27
+addons:
+  hosts:
+    - ip6-localhost
```
sibiryakov: Which address in `/etc/hosts` this hostname will be assigned to?

sibiryakov:
There is a ZMQ broker starting before tests. By default it's binding to `ZMQ_ADDRESS`, probably using IPv4. Would be nice to run it (for testing purposes) using wild-card.
I'm not sure that IPv6 test is actually testing IPv6 connectivity.

.travis.yml
lljrsr: That is a good question. The only thing I found when searching was [this](https://docs.travis-ci.com/user/hosts/)

sibiryakov: You could try adding/removing debug output to understand this.

lljrsr: Looks like it just assigns it to [127.0.0.1](https://travis-ci.org/scrapinghub/frontera/jobs/109851247#L121) :open_mouth: .. so this is probably not very helpful

lljrsr:
**Alright!**
Starting the broker with `--address='*'` now actually uses the wildcard and it is working. I double checked that in my distributed setup.

I included a few more brokers in the shell script (including IPv6) for testing. Travis is still complaining when testing IPv6 with `MessageBusTester()`. This is probably because the broker running in the background uses the default IPv4 config. Do you think there might be another issue causing this? Should I just remove the IPv6 test? ..testing with Travis is a pain and I did not properly set up my system for testing.. neither have I the time to do so.
It might be a good idea to create the broker during the test (not via shell script).. because there are different kinds of brokers now (This fix might take some time though which I do not have).

If you want to add the wildcard functionality also as a frontera setting for users I would ask you to do that in a different pull request because it might need some more rewriting. However I would not recommend this since only the broker can actually make use of this functionality (because `connect()` does not work with wildcards)

sibiryakov:
Yeah, let's disable IPv6 test for now, and create an issue that it was disabled, and needs to be enabled in a separate PR. It needs more investigation.

Agree about the broker, but I don't have clear idea how, yet.

I think that's ok to mention wild-card in settings docs (explaining that wild-card should be used only for brokers), and use `127.0.0.1` as default value.

Last thing, please rebase& squash your commits, there are two many of them. I think that's a good practice to do not trash the commits history.
Thanks for your work, as usual!

lljrsr:
Alright. Is it okay like this?

frontera/tests/run_zmq_broker.sh
```diff
@@ -1,2 +1,18 @@
 #!/usr/bin/env bash
-python -m frontera.contrib.messagebus.zeromq.broker 2> broker.log &
\ No newline at end of file
+# Start ZMQ broker with hostname instead of IP
+python -m frontera.contrib.messagebus.zeromq.broker --address localhost 2>> broker.log &
+sleep 5
```
sibiryakov: Wait, wait, wait :)
This is not reliable. Travis VMs are suffer from constantly changing performance. So, you never know for how long broker needs to be running.
Let's better run several broker instances using different ports till the end of test?

frontera/contrib/messagebus/zeromq/socket_config.py
```diff
@@ -1,29 +1,62 @@
...
-    hostname = None
-    base_port = None
-
-    def __init__(self, hostname, base_port):
-        self.hostname = hostname
-        self.base_port = base_port
+    """
+    Converts address to IPv4 or IPv6 or * and returns the necessary socket
+    addresses.
+    NOTE: When using * it defaults to IPv4
```
sibiryakov: This is an important detail. Let's move it to settings doc?

frontera/contrib/messagebus/zeromq/socket_config.py
```diff
@@ -1,29 +1,62 @@
...
-        return 'tcp://%s:%d' % (self.hostname, self.base_port + 3)
+        """
+        TCP socket for outgoing SW messages
+        """
+        return 'tcp://%s:%d' % (self.ip_addr, self.base_port + 3)
 
     def db_in(self):
-        return 'tcp://%s:%d' % (self.hostname, self.base_port + 4)
+        """
+        TCP socket for ingoing messages
```
sibiryakov: the right word would be "incoming"

sibiryakov:
Anyway, we're pretty close to merging.

frontera/contrib/messagebus/zeromq/socket_config.py
lljrsr: I only mentioned it in the code because IPv6 will probably not work properly (neither the test is implemented yet, nor was I able to try it out on my own machine).
But I will move it to the broker command line help and somewhere in the docs (I do not think it belongs to the `ZMQ_ADDRESS` setting since this setting is not supposed to be used with `*` anyway).
Do you want me to add a note saying that frontera has only been tested with IPv4?

sibiryakov: Man, if you were able to bind the socket on Travis CI on IPv6 address, then we have everything to make testing IPv6 work.
`ZMQ_ADDRESS` setting can be used with wildcard, it could be useful when running on different machines using config.
No, I won't surrender :) I'm going to find a machine in Scrapinghub, or setup a new one to test it in IPv6 environment.
If you don't want to continue, I'll take your code and finish it, just tell me.

lljrsr: Actually by using different ports for the broker, IPv6 testing should now be possible [see here](https://github.com/scrapinghub/frontera/pull/108/files#diff-d8026c0965fad80de128c149837278f9R81) (also the wildcard test could be added there).
I also updated some docs with my newest commit.
I do not think that I have the time tomorrow to make testing of IPv6 and wildcard work on Travis. But even if Travis testing worked it would only test this on `ip6-localhost`.. not in a distributed setup. So I could not say for sure if IPv6 would work.
Of course after friday you can happily take over. In my opinion IPv6 support could still be in a separate pull request since my changes will not break anything.

frontera/tests/run_zmq_broker.sh
```diff
@@ -1,2 +1,15 @@
 #!/usr/bin/env bash
-python -m frontera.contrib.messagebus.zeromq.broker 2> broker.log &
\ No newline at end of file
+# Start ZMQ broker with hostname instead of IP
+python -m frontera.contrib.messagebus.zeromq.broker --address localhost --port '5580' 2>> broker.log &
+kill $(ps aux | grep 'python -m frontera.contrib.messagebus.zeromq.broker' | awk '{print $2}')
+
+# Start ZMQ broker with IPv6
+python -m frontera.contrib.messagebus.zeromq.broker --address '::1' --port '5570' 2>> broker.log &
+kill $(ps aux | grep 'python -m frontera.contrib.messagebus.zeromq.broker' | awk '{print $2}')
```
sibiryakov: why do you kill broker after running it?

lljrsr: Don't ask :P .

lljrsr:
Do you have anything else to add?
I only have a few hours left ;) .

sibiryakov:
Awesome! Thank you! And good luck at the new place!

Issue #116, opened 2016-02-26
Link to "Frontier at a glance" on the docs is broken


josericardo:
The `Frontier at a glance` link here

http://frontera.readthedocs.org/en/latest/topics/scrapy-integration.html#spider-logic

is broken.

sibiryakov:
Fixed!

PR #119, opened 2016-03-17
fix for issue #116

Vlsarro:
https://github.com/scrapinghub/frontera/issues/116
'frontier at a glance' page is almost identical to 'Quick start single process'. So just changed that link. You can compare the following links and see for yourself.
http://frontera.readthedocs.org/en/latest/topics/quick-start-single.html
http://frontera.readthedocs.org/en/v0.3.0/topics/frontier-at-a-glance.html (found only v0.3.0 docs)


sibiryakov:
Thank you :beer:

PR #122, opened 2016-03-25
make snappy codec optional [Work in progress]

jp111:
I have made the snappy codec optional, I am not sure how to remove it from requirements.


sibiryakov:
That will not work. You would also have to update the class instantiating producers.

sibiryakov:
@jp111 any progress on this?

PR #133, opened 2016-04-09
utils.async unit tests and minor fix

Preetwinder:
I have added some tests for the two functions in utils.async. A small modification to listen_tcp - I have changed the check `not list` to `is int` because this is more accurate as the argument to `listenTCP` must be integral. Also I have moved down the assert statement because otherwise `len` will fail on integer arguments, thus the `is int` if statement won't be reached.
I have used [Clock](https://twistedmatrix.com/documents/current/api/twisted.internet.task.Clock.html) to mock the reactor for `CallLaterOnce`, and I have used [MemoryReactor](https://twistedmatrix.com/documents/current/api/twisted.test.proto_helpers.MemoryReactor.html) to mock the reactor for `listen_tcp`.

### DIFF
--- a/frontera/tests/test_utils_async.py
+++ b/frontera/tests/test_utils_async.py
@@ -58,7 +58,8 @@ class TestListenTCP(object):
 
     def test_listen_tcp_integer(self):
         reactor = MemoryReactor()
-        listen_tcp(self.port, self.host, Factory, reactor=reactor)
+        result = listen_tcp(self.port, self.host, Factory, reactor=reactor)
+        assert result.getHost().port == self.port
         assert reactor.tcpServers[0][0] == self.port
 
     def test_listen_tcp_invalid_port_range(self):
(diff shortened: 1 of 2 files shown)


sibiryakov:
Hey Preet,
Tanks for the PR. That looks interesting, I see you put some thought into it. Let me look through it more deeply tomorrow. 

Отправлено с iPhone

frontera/tests/test_utils_async.py
```diff
@@ -0,0 +1,84 @@
...
+    called = 0
+
+    def call_function(self):
+        self.called += 1
+
+    def test_call_later_once(self):
+        self.called = 0
+        reactor = Clock()
+        call = CallLaterOnce(self.call_function, reactor=reactor)
+        call.schedule(delay=1)
```
sibiryakov: can we decrease delays here to millisecond? Tests have to run fast.

frontera/tests/test_utils_async.py
```diff
@@ -0,0 +1,84 @@
...
+
+class TestListenTCP(object):
+
+    host = '127.0.0.1'
+    port = 6023
+    portrange = [6023, 6073]
+
+    def test_listen_tcp_integer(self):
+        reactor = MemoryReactor()
+        listen_tcp(self.port, self.host, Factory, reactor=reactor)
```
sibiryakov: There is a return value, which also can be asserted.

sibiryakov:
So far your code looks good!

Preetwinder:
Thanks. I have added asserts for returned values. The delay can be decreased, but it isn't a problem since there is no actual reactor running so there is no delay.

sibiryakov:
Thank you! 👍

PR #134, opened 2016-04-13
SQLA unicode issues

sibiryakov:
Sometimes SQLA returns strings as unicode, and it brakes Kafka client.


sibiryakov:
so we don't want fingerprint or method to be in unicode, but we're ok if URL will be unicoded.

sibiryakov:
It looks like SQLA returns unicode for strings depending on internal backend to support unicode. For example sqlite stores strings internally as unicode, and returns them as unicode through python binding. The same isn't true for MySQL for example.

PR #141, opened 2016-05-12
wrong number

happywindw:
a little mistake

### DIFF
--- a/docs/source/topics/production-broad-crawling.rst
+++ b/docs/source/topics/production-broad-crawling.rst
@@ -53,7 +53,7 @@ Strategy worker is reading :term:`spider log`, calculating score, deciding if UR
 update_score events to :term:`scoring log`.
 
 Before setting it up you have to decide how many spider instances you need. One spider is able to download and parse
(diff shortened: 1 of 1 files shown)


sibiryakov:
@happywindw I see no mistake here. I managed to get 1200 pages/minute with Scrapy, but I put there half of that, because speed is highly dependent on parsing operation CPU consumption, internet connectivity and various Scrapy settings.

happywindw:
Sorry, my mistake by seen pages/minute as pages/second, and thanks for your reply.
I understand that speed dependent on many factors. But in the doc 700 pages/minute is about 12 pages/second, so 10 spiders is about 120 pages/second or less, that is far away from 1K/second, right?

sibiryakov:
It's not necessary you will get 700 p/min. I would recommend to measure performance and make assumptions based on your measurements. A simple spider which parses HTML using lxml and extracts links from it was making 1200 pages/minute in average. Which is 20 p/sec and you need 50 spiders to get to 1K/second. Which is about 7 machines with 8 cores each.

happywindw:
ok, got it. thanks again!

PR #147, opened 2016-05-27
Tests fort utils.url

Preetwinder:
The tests for `safe_url_string` and `canonicalize_url` were taken from the older versions of scrapy and w3lib, since the newer versions of these functions are backwards incompatible.


sibiryakov:
@Preetwinder for the record, what are incompatibilities?
Overall, these tests do require some competence in this problem. I feel that I'm lack of canonicalization concept in Scrapy, let's wait till @redapple get back from his vacation.

sibiryakov:
It seems fine!

PR #150, opened 2016-06-06
Tests for frontera scheduler

Preetwinder:
Both the crawler and FrontierManager are faked. The fake crawler is needed to calculate the overused keys calculation, so it only contains some settings, and a function to setup fake slot info. The fake FrontierManager records all the requests, responses, errors, and arguments, in lists. The actual test just call the functions, and verify the types and content of the values stored in the FrontierManager object.

### DIFF
--- a/frontera/tests/test_frontera_scheduler.py
+++ b/frontera/tests/test_frontera_scheduler.py
@@ -1,0 +1,150 @@
+from frontera.contrib.scrapy.schedulers.frontier import FronteraScheduler
+from frontera.tests.mocks.frontier_manager import FakeFrontierManager
+from frontera.tests.mocks.crawler import FakeCrawler
+from frontera.core.models import Request as FRequest
+from frontera.core.models import Response as FResponse
+from scrapy.http import Request, Response
+from scrapy.spiders import Spider
+from scrapy.settings import Settings
+
+
+# test requests
+r1 = Request('http://www.example.com')
+r2 = Request('https://www.example.com/some/page')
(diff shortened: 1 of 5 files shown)


frontera/tests/test_frontera_scheduler.py
```diff
@@ -0,0 +1,150 @@
...
+        fs.open(Spider)
+        fs.frontier.manager.put_requests([fr1])
+        fs.frontier.manager.finished = True
+        assert fs.next_request() is None
+        assert fs.stats_manager.stats.get_value('frontera/returned_requests_count') is None
+
+    def test_next_request_overused_keys_info(self):
+        settings = Settings()
+        settings['CONCURRENT_REQUESTS_PER_DOMAIN'] = 0
+        settings['CONCURRENT_REQUESTS_PER_IP'] = 1
```
sibiryakov: It would be nice to test it with bigger values, say 5. With 1 you're limiting it's functionality.

sibiryakov:
Overall, it's wonderful! Keep going, @Preetwinder.

frontera/tests/test_frontera_scheduler.py
Preetwinder: I have made this change, although since we are using a fake downloader, other changes are not needed in the test to make it work.

sibiryakov:
I think we could continue with that. In real downloader requests are processed in nearly random order, and I think such behavior needs to be simulated in tests. But let's leave it for now, we have a lot of things to do.

PR #151, opened 2016-06-06
Tests for frontier manager

Preetwinder:
Fake backend, canonicalsolver and middlewares are used. There are a few different middlewares, some don't pass the values through, some modify some values etc. Each of these component records all the requests, responses, errors in lists. In the tests we verify the type and content of values passed through, and whether they were modified.

### DIFF
--- a/frontera/tests/mocks/components.py
+++ b/frontera/tests/mocks/components.py
@@ -1,0 +1,149 @@
+from frontera.core.components import Backend, Middleware, CanonicalSolver
+
+
+class FakeBackend(Backend):
+
+    def __init__(self, manager):
+        self.seeds = []
+        self.links = []
+        self.responses = []
+        self.requests = []
+        self.errors = []
+        self.lists = [self.seeds, self.responses, self.links, self.errors]
+        self._started = False
+        self._stopped = False
+        self._finished = False
+
+    @classmethod
+    def from_manager(cls, manager):
+        return cls(manager)
+
+    def frontier_start(self):
+        self._started = True
+
+    def frontier_stop(self):
+        self._stopped = True
+
+    def finished(self):
+        return self._finished
+
+    def add_seeds(self, seeds):
+        for seed in seeds:
+            self.seeds.append(seed)
+
+    def page_crawled(self, response, links):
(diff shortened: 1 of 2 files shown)


frontera/tests/mocks/components.py
```diff
@@ -0,0 +1,149 @@
...
+        for i in range(max_next_requests):
+            if self.requests:
+                lst.append(self.requests.pop())
+        return lst
+
+    def request_error(self, request, error):
+        self.errors.append((request, error))
+
+
+class FakeMiddleware(Middleware):
```
sibiryakov: Is it possible to unify FakeMiddleware and FakeBackend? They look quite similar.

frontera/tests/test_frontier_manager.py
```diff
@@ -0,0 +1,111 @@
...
+        assert set([link for link in fm.backend.links]) == set([r2, r3])
+        assert set([link for link in fm.canonicalsolver.links]) == set([r2, r3])
+        assert [set([link for link in mw.links]) for mw in fm.middlewares] == [set([r2, r3])]*4
+        assert [link.meta['test_links'] for link in [r2, r3]] == ['test']*2
+        assert [link.meta['test_links_canonical_solver'] for link in [r2, r3]] == ['test']*2
+
+    def test_get_next_requests(self):
+        fm = self.setup_frontier_manager()
+        fm.backend.put_requests([r1, r2, r3])
+        assert set(fm.get_next_requests(3)) == set([r1, r2, r3])
```
sibiryakov: what if we add here asserts for iteration and n_requests?

sibiryakov:
Let's add test for finishing condition implemented in Backend also. 
In general, it looks great! I like it.

frontera/tests/mocks/components.py
Preetwinder: I have made the FakeBackend a subclass of the FakeMiddleware.

Preetwinder:
I have added a test for max_requests reached, and also added checks for iteration and n_requests.

sibiryakov:
Looks great!

PR #161, opened 2016-06-20
Fix SQL integer type for crc32 field

lopuhin:
CRC32 is an unsigned 4-byte int, so it does not fit in a signed 4-byte int (``Integer``). There is no unsigned int type in the SQL standard, so I changed it to ``BigInteger `` instead. Without this change, both MySQL and Postgres complain that ``host_crc32`` field value is out of bounds. 
Another option (to save space) would be to conver CRC32 into a signed 4-bit int, but this will complicate things, not sure it's worth it.

### DIFF
--- a/tests/test_utils_misc.py
+++ b/tests/test_utils_misc.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
+import hashlib
 import pytest
-from frontera.utils.misc import load_object, get_crc32, chunks
+from frontera.utils.misc import load_object, get_crc32, chunks, to_signed32
 import six
 
 
@@ -17,6 +18,17 @@ class TestGetCRC32(object):
 
     def test_non_ascii_bytes(self):
         assert get_crc32(u'example\u5000'.encode('utf8')) == 1259721235
+
+    def test_negative_crc32(self):
+        assert get_crc32(b'1') == -2082672713
+
+    def test_crc32_range(self):
+        left, right = -2**31, 2**31 - 1
+        for x in range(10000):
+            assert left <= get_crc32(hashlib.md5(str(x)).hexdigest()) <= right
+        for x in [left, left + 1, right - 1, right, right + 1,
+                  2**32 - 2, 2**32 - 1]:
+            assert left <= to_signed32(x) <= right
 
 
 class TestChunks(object):
--- a/tests/test_utils_misc.py
+++ b/tests/test_utils_misc.py
@@ -25,7 +25,8 @@ class TestGetCRC32(object):
     def test_crc32_range(self):
         left, right = -2**31, 2**31 - 1
         for x in range(10000):
-            assert left <= get_crc32(hashlib.md5(str(x)).hexdigest()) <= right
+            bytestr = hashlib.md5(str(x).encode('ascii')).hexdigest()
+            assert left <= get_crc32(bytestr) <= right
         for x in [left, left + 1, right - 1, right, right + 1,
                   2**32 - 2, 2**32 - 1]:
             assert left <= to_signed32(x) <= right
--- a/frontera/contrib/backends/partitioners.py
+++ b/frontera/contrib/backends/partitioners.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
-from zlib import crc32
 from struct import unpack
 from binascii import unhexlify
 
--- a/frontera/utils/misc.py
+++ b/frontera/utils/misc.py
@@ -32,7 +32,19 @@ def load_object(path):
 
 
 def get_crc32(name):
-    return crc32(to_bytes(name, 'utf-8', 'ignore'))
+    """ signed crc32 of bytes or unicode.
+    In python 3, return the same number as in python 2, converting to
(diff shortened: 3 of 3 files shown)


sibiryakov:
It's not always unsigned ;)
https://docs.python.org/2/library/zlib.html?highlight=crc32#zlib.crc32
What if we go for recommended ``crc32(data) & 0xffffffff`` approach?

MySQL has unsigned int, but Postgres doesn't. I guess you're right, BIGINT would be an easiest way in the age of cheap hardware. It would be nice to write a test case for that. We do have a checks in Frontera for MySQL, sqlite and Postgresql.

lopuhin:
Ah, you are right, in fact it's a python3-only issue - I did not realize that crc32 gives different results in py2 and py3 :)

Since it's already signed in python 2, and we'll need some compatibility logic for conversion anyway, maybe it's better to convert it so a signed int on py3? This way we will remain backwards compatible (both in data and in the table schema), and will not waste space for storage. What do you think?

@sibiryakov could you point me where the different backend tests are located?

sibiryakov:
We claim support for Python 2.7+, so I think that's fine to convert it to signed int. That should work on all platforms and CPython versions.

sibiryakov:
@lopuhin any update on this?

sibiryakov:
@lopuhin Костя, давай уже что-нибудь сделаем :)

lopuhin:
@sibiryakov oops, sorry, I postponed it for too long! Let me do it today :)

lopuhin:
I've converted all usages of ``binascii.crc32`` to ``frontera.utils.misc.get_crc32``, and updated the ``get_crc32`` function to return the same result on both python 2 and python 3 (without that change, ``get_crc32(b'1')`` would return 2212294583 on python 3), so the new test will pass on python 3 too.

sibiryakov:
@lopuhin it would be nice to test that it doesn't crosses 4-byte signed integer border. Also, PR needs rebasing, since we're working on PY3 port ;)

lopuhin:
@sibiryakov right, the range check is crucial here, I rebased and added the check. Py3 port is great news!

tests/test_utils_misc.py
```diff
@@ -18,6 +19,17 @@ def test_non_ascii_unicode(self):
     def test_non_ascii_bytes(self):
         assert get_crc32(u'example\u5000'.encode('utf8')) == 1259721235
 
+    def test_negative_crc32(self):
+        assert get_crc32(b'1') == -2082672713
+
+    def test_crc32_range(self):
```
sibiryakov: Is this test fails when get_crc32 used without to_signed32? Sorry, for being picky @lopuhin

lopuhin: Np! :)
Yep, both these tests fail without ``to_signed32`` under py35 for me (I fixed the py35 test in 4a10c08, didn't notice it was failing at first).
```
(py35) frontera (git::sql-crc3-int-type) $ git diff
diff --git a/frontera/utils/misc.py b/frontera/utils/misc.py
index f7372bc..b1c7933 100644
--- a/frontera/utils/misc.py
+++ b/frontera/utils/misc.py
@@ -44,6 +44,7 @@ def get_crc32(name):
 def to_signed32(x):
     """ If x is an usigned 32-bit int, convert it to a signed 32-bit.
     """
+    return x
     return x - 0x100000000 if x > 0x7fffffff else x
 
 
(py35) frontera (git::sql-crc3-int-type) $ py.test tests/test_utils_misc.py::TestGetCRC32
============================================================================== test session starts ===============================================================================
platform darwin -- Python 3.5.1, pytest-2.9.2, py-1.4.31, pluggy-0.3.1
rootdir: /Users/kostia/shub/frontera, inifile: tox.ini
collected 6 items 

tests/test_utils_misc.py ....FF
============================================================================ short test summary info =============================================================================
FAIL tests/test_utils_misc.py::TestGetCRC32::()::test_negative_crc32
FAIL tests/test_utils_misc.py::TestGetCRC32::()::test_crc32_range

==================================================================================== FAILURES ====================================================================================
________________________________________________________________________ TestGetCRC32.test_negative_crc32 ________________________________________________________________________

self = <tests.test_utils_misc.TestGetCRC32 object at 0x105e20d30>

    def test_negative_crc32(self):
>       assert get_crc32(b'1') == -2082672713
E       assert 2212294583 == -2082672713
E        +  where 2212294583 = get_crc32(b'1')

tests/test_utils_misc.py:23: AssertionError
_________________________________________________________________________ TestGetCRC32.test_crc32_range __________________________________________________________________________

self = <tests.test_utils_misc.TestGetCRC32 object at 0x105e17e80>

    def test_crc32_range(self):
        left, right = -2**31, 2**31 - 1
        for x in range(10000):
            bytestr = hashlib.md5(str(x).encode('ascii')).hexdigest()
>           assert left <= get_crc32(bytestr) <= right
E           assert 2157876746 <= 2147483647
E            +  where 2157876746 = get_crc32('a87ff679a2f3e71d9181a67b7542122c')

tests/test_utils_misc.py:29: AssertionError
======================================================================= 2 failed, 4 passed in 0.12 seconds =======================================================================
```

sibiryakov:
LGTM. @Preetwinder ?

Preetwinder:
+1. Looks good.

PR #168, opened 2016-06-27
PY3 Syntactic changes.

Preetwinder:
Most of the changes were produced using the modernize script. Changes include print syntax, error syntax, converting iterators and generators to lists, etc. Also includes some other changes which were missed by the script.


redapple:
hi @Preetwinder , what do you think of https://github.com/scrapinghub/frontera/pull/166 ?
I'd like to get code coverage running on Travis so that we can more easily see where tests are further needed.
If we're ok, this PR could probably be rebased on it easily.
cc @sibiryakov

Preetwinder:
Hello. Yes I think the changes in #166 are great. Judging from the stats on the codecov site adding tests for hbase, kafka and workers should give us a reasonably good coverage. I'll rebase my branch once that PR is merged(or do I need to do it before?). Also should I wait for that PR to be merged before making my other PR's?

redapple:
@Preetwinder , you may be touching a few of the same files as #166 next, so I'm not sure if you should change too much "by hand". If most of this PR has been done with scripts it may not be too hard to fix any rebase issue.
But ideally, #166 would be merged before further change (so that we track code coverage PR by PR)
I believe @sibiryakov could approve (or not) #166 soon enough

redapple:
@Preetwinder , #166 has been merged.
Would you mind rebasing against current master branch?

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -4,12 +4,14 @@
...
 from frontera.core.codec import BaseDecoder, BaseEncoder
+import six
+from six.moves import map
 
 
 def _prepare_request_message(request):
     def serialize(obj):
         """Recursively walk object's hierarchy."""
-        if isinstance(obj, (bool, int, long, float, basestring)):
+        if isinstance(obj, (bool, int, float, six.string_types)):
```
redapple: > if isinstance(obj, (bool, int, long, float, basestring)):
is there a case where a `long` is passed or is this theoretical?

frontera/core/codec.py
```diff
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import
 from abc import ABCMeta, abstractmethod
+import six
 
 
-class BaseDecoder(object):
-    __metaclass__ = ABCMeta
-
+class BaseDecoder(six.with_metaclass(ABCMeta, object)):
```
redapple: I also like the [`add_metaclass` decorator way](https://pythonhosted.org/six/#six.add_metaclass) but it's nitpicking :)

redapple:
Overall, the changes look good to me.
It's true there are a lot of `.iteritems()` wrapped in `six.iteritems()` now. I'm not sure all cases really need the iterator version in Python 2 (so that using `.items()` without `six` would be good enough in both Py2 and Py3) but it would be very long to check.

frontera/contrib/backends/remote/codecs/msgpack.py
Preetwinder: Yes I think this case should be covered. I have used six.integer_types instead. It covers both `long` and `int` type.

frontera/core/codec.py
Preetwinder: I agree. It looks better.

redapple:
LGTM

sibiryakov:
Hey @Preetwinder and @redapple in order to merge this we should enable tests for Python 3 also, otherwise how would we find out if it works. Next thing to do is, to port Kafka message bus testing from https://github.com/scrapinghub/frontera/blob/145d0b58da0ce5175ba2795ea76f63f1e2a12184/frontera/tests/test_message_bus.py

and test HBase stuff at least manually.

redapple:
@Preetwinder , @sibiryakov ,
so I tried with Python 3 on Travis CI (see https://github.com/scrapinghub/frontera/issues/175 , https://travis-ci.org/scrapinghub/frontera/builds/145764784 )
but MySQL-python is having trouble getting installed.

`requirements/tests.txt` also includes https://github.com/PyMySQL/PyMySQL
where/when is MySQL-python>=1.2.5 needed?

Preetwinder:
It seems MySQL-python is used by default in the MySQL tests. Since it doesn't support Python 3, I think we can force the tests to use PyMySQL instead and remove MySQL-python.

redapple:
SQLAchemy mentions https://github.com/PyMySQL/mysqlclient-python
http://docs.sqlalchemy.org/en/latest/dialects/mysql.html#py3k-support

redapple:
@Preetwinder , Python 3 tests with PyMySQL fail badly (dont know why)
It may be easier to use `mysqlclient-python` (Python3 compatible) instead of forcing `+pymysql` (worth a Travis test build at least I think)

Preetwinder:
I'll try using mysqlclient. I think the reason ZeroMQ tests fail might be that since both the PY2 and PY3 tests are being run in the Python 3 environment, this file - https://github.com/scrapinghub/frontera/blob/master/tests/run_zmq_broker.sh uses Python 3 even for the Python 2 tests.

redapple:
I've killed the Py35 build on Travis

redapple:
@Preetwinder , tests are run within a virtualenv thanks to tox.
`run_zmq_broker.sh` should also run inside the virtualenv (this ought to be possible with tox I believe)

sibiryakov:
@Preetwinder Please try running pyMysql locally, it should be clearer why it looses connection constantly. I wonder if it connects at all.

Preetwinder:
In order to make pymysql work I have pushed most of the PY3 modification for single process mode here(I was originally planning to do it in a separate PR). I'll continue testing and improving the PY3 changes. Currently message_bus and scrapy_spider tests fail for PY2, I'll look into this, although the cause might be unrelated to this PR(they fail for me locally even without the changes in this PR). For PY3 the failures for message_bus, canonicalize_url and encoders are expected since I haven't worked on these modules yet. There is also an import error for sgmllib which seems to be a deprecated module.

redapple:
Regarding `sgmllib`, indeed, `RegexLinkExtractor` extends `SgmlLinkExtractor` -- which has been deprecated in Scrapy, and is not available in Python 3. (This makes me think that Scrapy could ship with a robust regex link extractor, without the `sgmllib` dependency, which doesn't bring much (anything) to `RegexLinkExtractor` implementation -- There is even [a comment about this in scrapy tests](https://github.com/scrapy/scrapy/blob/ff80e1c38110a6105db5d5e34570de8c9ece4982/tests/test_linkextractors_deprecated.py#L177))

Preetwinder:
The reason for message_bus test failure on PY2 was the tests being run with language set as PY3.5. The scrapy_spider tests was failing because the test spider would be unable to crawl past the login page on the scrapinghub website. I have changed the website to dmoz.org.

tests/test_scrapy_spider.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import
```
redapple: 2 options for this test:

- regarding `sgmllib` not available in Python 3, I suggest you add a _skip in Python 3_ for  `test_scrapy_spider`
- or remove `RegexLinkExtractor` from https://github.com/scrapinghub/frontera/blob/master/tests/scrapy_spider/spiders/example.py#L23

sibiryakov:
@redapple and @Preetwinder Let's better remove RegexLinkExtractor. The goal of this test is to check the general functionality - if all is imported well, and Scrapy/Frontera API's fit.

redapple:
Regarding 
```
FAIL tests/test_utils_url.py::TestCanonicalizeUrl::test_non_ascii_percent_encoding_in_path
FAIL tests/test_utils_url.py::TestCanonicalizeUrl::test_non_ascii_percent_encoding_in_query_argument
FAIL tests/test_utils_url.py::TestCanonicalizeUrl::test_normalize_percent_encoding_in_path
FAIL tests/test_utils_url.py::TestCanonicalizeUrl::test_normalize_percent_encoding_in_query_arguments
FAIL tests/test_utils_url.py::TestCanonicalizeUrl::test_safe_characters_unicode
```
`canonicalize_url` implementation should match scrapy's (which I believe it correct), and tests need to be changed accordingly

sibiryakov:
@redapple what if we move that part to standalone library? How time consuming is it?

redapple:
@sibiryakov , you mean in w3lib?

sibiryakov:
May be! Consider also cython compilation in the feature.

redapple:
@Preetwinder , what do you think on @sibiryakov comments on removing `RegexLinkExtractor`?
I doubt we'll have `canonicalize_url` in w3lib shortly, so the quickest here is to copy-paste the implementation in frontera.

Preetwinder:
Yes I agree with @sibiryakov. I'll remove it. I can copy-paste the function and it's tests from scrapy, but the issue about this was that changing the implementation of `canonicalize_url` changes the fingerprint calculated in the URL middleware. If you and @sibiryakov are fine with this change, I'll make it.

redapple:
IMO, 'old' `canonicalize_url` was broken, so some fingerprints will be invalidated with this, but it's for the better.

redapple:
@Preetwinder , what about you submit a PR to w3lib about adding `canonicalize_url()`?
There's already an issue for it: https://github.com/scrapy/w3lib/issues/65

Removing `canonicalize_url()` from scrapy can be done at later stage.

sibiryakov:
@Preetwinder I'm fine with this change. But [dramatic pause] we should mention that in release notes!

frontera/contrib/backends/memory/__init__.py
```diff
@@ -128,11 +135,11 @@ def _get(self, obj):
 
     def update_cache(self, objs):
         objs = objs if type(objs) in [list, tuple] else [objs]
-        map(self._put, objs)
+        list(map(self._put, objs))
```
sibiryakov: Why list is needed here? The idea is to apply self._put to every element in the objs

Preetwinder: I think it is because map returns a generator in PY3. So if we don't convert it to a list, it never get's applied.

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -80,18 +82,18 @@ def decode(self, buffer):
         if obj[0] == 'pc':
             return ('page_crawled',
                     self._response_from_object(obj[1]),
-                    map(self._request_from_object, obj[2]))
+                    list(map(self._request_from_object, obj[2])))
```
sibiryakov: Agreed privately, that Preet will check all cases where list is used, and remove those where it's not needed.

frontera/contrib/backends/remote/kafka.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
```
sibiryakov: This whole module is obsolete and can be removed in favor of Kafka message bus.

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -79,7 +83,7 @@ def _modify_page(self, obj):
     def _create_page(self, obj):
         db_page = self.model()
         db_page.fingerprint = obj.meta['fingerprint']
-        db_page.url = obj.url
+        db_page.url = obj.url_as_unicode
```
sibiryakov: why it's converted here to unicode? May be we discussed that, but I please help me to remember the reasoning.

Preetwinder: We haven't discussed this. I did this because by converting the url to unicode, I was able to avoid some encoding issues with some of the DB's. Non-ascii bytes were causing some issues for some of the them.

frontera/contrib/backends/remote/kafka.py
Preetwinder: Should I remove it in a seperate PR?

frontera/contrib/scrapy/converters.py
```diff
@@ -98,8 +99,8 @@ def from_frontier(self, response):
 
 
 def _find_method(obj, func):
-    if obj and hasattr(func, 'im_self') and func.im_self is obj:
-        return func.im_func.__name__
+    if obj and hasattr(func, '__self__') and func.__self__ is obj:
```
sibiryakov: are you sure that would work the same way in py2.7?

frontera/contrib/scrapy/middlewares/seeds/__init__.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
```
sibiryakov: is absolute import needed here?

frontera/core/models.py
```diff
@@ -111,6 +140,27 @@ def url(self):
...
+
+    @property
+    def url_as_bytes(self):
+        """
+        Returns the URL as bytes.
+        """
+        return to_bytes(self.url, self.encoding)
+
+    @property
+    def encoding(self):
```
sibiryakov: Are codecs and other serialization methods support storing of encoding?

frontera/contrib/scrapy/converters.py
Preetwinder: According to the second to last bullet point under this heading - https://docs.python.org/dev/whatsnew/2.6.html#other-language-changes `__self__` was added as an alternative name in PY2.6. So this should work for PY3 and >=2.6.

frontera/settings/default_settings.py
```diff
@@ -50,6 +51,7 @@
 }
 SQLALCHEMYBACKEND_REVISIT_INTERVAL = timedelta(days=1)
 STATE_CACHE_SIZE = 1000000
+STATE_CACHE_SIZE_LIMIT = 0
```
sibiryakov: Where is this comes from? I see it's not used anywhere in this PR.

frontera/contrib/scrapy/middlewares/seeds/__init__.py
Preetwinder: A lot of the usage of absolute_import is preemptive. So it might not be needed for a lot of the cases. If you think it will affect performance, the one's which aren't needed can be removed.

frontera/core/models.py
Preetwinder: I haven't made the decision about whether the encoding needs to be stored. So right now the encoding is only stored in the Request/Response objects. I did this because we don't need to make this decision for single process mode's working.

frontera/settings/default_settings.py
Preetwinder: The reasoning for this is that if we access `STATE_CACHE_SIZE_LIMIT` without specifying a value. We get None. Comparing integers with None works in PY2, but is an error in PY3. So https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/memory/__init__.py#L141 would fail.

frontera/utils/misc.py
```diff
@@ -29,11 +32,11 @@ def load_object(path):
 
 
 def get_crc32(name):
-    return crc32(name) if type(name) is str else crc32(name.encode('utf-8', 'ignore'))
+    return crc32(to_bytes(name, 'utf-8', 'ignore'))
 
 
 def chunks(l, n):
-    for i in xrange(0, len(l), n):
+    for i in range(0, len(l), n):
```
sibiryakov: This is slower for Py2.7, and I'm not sure about Py3. What if we could try to achieve the same behavior by returning a list of iterators created by itertools.islice?

sibiryakov:
@Preetwinder please avoid such a big pull requests next time, it's extremely hard to review and discuss it.

frontera/utils/misc.py
redapple: In https://pythonhosted.org/six/, it seems that in Python 2, `six.moves.range` uses `xrange`, so I'm not sure it's slower here.

redapple: There's also the [grouper recipe from itertools](https://docs.python.org/3/library/itertools.html#itertools-recipes)
```
def grouper(iterable, n, fillvalue=None):
    "Collect data into fixed-length chunks or blocks"
    # grouper('ABCDEFG', 3, 'x') --> ABC DEF Gxx"
    args = [iter(iterable)] * n
    return zip_longest(*args, fillvalue=fillvalue)
```

frontera/contrib/backends/memory/__init__.py
sibiryakov: the map was used here, because in Py2.7 it's faster than iterate with for cycle. According to my tests the 
```python
[f(x) for x in data] # is faster than

list(map(f, data))
```

in python 3.5
So I'm suggesting to rewrite it that way.

frontera/contrib/backends/remote/kafka.py
sibiryakov: Yes.

frontera/contrib/backends/sqlalchemy/components.py
sibiryakov: We're building a large scale crawling software, and we have to save resources: processor time, storage capacity, network throughput. Before adding a new field, which will consume all these resources we need to have a strong reason why we're doing that. Talking with @redapple in private Slack we didn't found such a strong reason.

In fact Paul thinks that unicode URLs are only needed if they are going to be reviewed by human. In Frontera, this is quite rare.

He suggested to use https://github.com/scrapy/scrapy/blob/master/scrapy/http/request/__init__.py#L53 safe_url_string from w3lib to convert all URLs to ASCII. 

Let's agree, that in Frontera we store and transfer URLs in ASCII encoding. If we receive them in unicode or other encoding we convert them to ASCII.

frontera/contrib/scrapy/converters.py
sibiryakov: That's fine. Is there tests for serializing/unserializing requests with callback function?

frontera/contrib/scrapy/middlewares/seeds/__init__.py
sibiryakov: I think it is a useless code, and that's why it should be removed.

frontera/core/models.py
sibiryakov: I would remove encoding from these objects, please read this comment https://github.com/scrapinghub/frontera/pull/168#discussion_r72272128

frontera/settings/default_settings.py
sibiryakov: Ok.

frontera/contrib/scrapy/middlewares/seeds/__init__.py
Preetwinder: Porting guides mostly recommend making this change for pragmatic reasons - [1 ](https://docs.python.org/3/howto/pyporting.html#prevent-compatibility-regressions),   [2 ](https://wiki.python.org/moin/PortingPythonToPy3k),   [3 ](http://python-future.org/compatible_idioms.html#imports-relative-to-a-package).
If you think it is adding needless noise, most of these statements can be removed without any affect(just a final confirmation from you).

frontera/contrib/scrapy/converters.py
Preetwinder: There is a test for to_frontier conversion - https://github.com/scrapinghub/frontera/blob/master/tests/test_scrapy.py#L34
In the latest commit I have added a test for from_frontier conversion.

frontera/utils/misc.py
Preetwinder: I tested a version based on `islice`, but it is slower. The grouper recipe shared by redapple is faster than the existing function but it introduces `None` values into the returned lists when they are not split evenly(example - `[1,2,3,4,5] -> [[1,2], [3,4],[5, None]])`. So we need to change the way the function is used in frontera(this will decrease the overall performance gain).

Preetwinder:
@sibiryakov I apologize. I wasn't planning to make all the changes in the same PR(hence the name of the PR), but you can see the series of events above which led me to do this.

frontera/core/models.py
```diff
@@ -23,10 +24,7 @@ def __init__(self, url, method='GET', headers=None, cookies=None, meta=None, bod
         :param dict cookies: dictionary of cookies to attach to this request.
         :param dict meta: dictionary that contains arbitrary metadata for this request.
         """
-        self._url = url
-        self._method = str(method).upper()
-        self._url = to_native_str(url, encoding)
-        self._encoding = encoding
+        self._url = safe_url_string(url, encoding)
```
sibiryakov: This is very expensive, comparing to what we have now...

sibiryakov:
@Preetwinder I would propose to collect the cases where we slowed down Frontera in a separate issue, and address that issue after we finish working on PY3 port. It can be seen as lending performance debt, and then giving ti back.

frontera/core/models.py
Preetwinder: This change isn't necessary to use frontera with scrapy, since scrapy already converts URL's to safe_url's. I'll remove it for now.

frontera/contrib/scrapy/middlewares/seeds/__init__.py
sibiryakov: Ok, let's leave them for pragmatic reasons.

frontera/core/models.py
sibiryakov: Ideal would be to add additional field for storing parsed URL. Because many components are performing url parsing, so they could make use of such structure.

sibiryakov: Frontera is designed to work without Scrapy too. So, we need this URL transformation, the question is how to make it cheap. That's why I propose to optimize it later.

frontera/contrib/backends/memory/__init__.py
```diff
@@ -1,14 +1,21 @@
...
 from frontera.core import OverusedBuffer
 from frontera.utils.heap import Heap
 from frontera.contrib.backends.partitioners import Crc32NamePartitioner
 from frontera.utils.url import parse_domain_from_url_fast
+import six
+from six.moves import map
+from six.moves import range
+
+
+def cmp(a, b):
```
sibiryakov: Is this function used somewhere?

frontera/settings/default_settings.py
sibiryakov: I still don't see why state_cache_size_limit option was introduced and where it is used?

frontera/contrib/backends/memory/__init__.py
Preetwinder: In python 3 cmp is not an in-built function. It is used here - https://github.com/Preetwinder/frontera/blob/ec0811a21983ab724948ccb7a9e8fdea759e42b9/frontera/contrib/backends/memory/__init__.py#L80 and a few other locations below this line.

frontera/settings/default_settings.py
Preetwinder: It is accessed here - https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/sqlalchemy/__init__.py#L41. Then it is passed to SQLACHEMY State object here - https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/sqlalchemy/components.py#L108. Then is passed to MemoryState via super and then the problem in the original comment arises.

Preetwinder: Do you think we should use a different value other than 0? For hbase it is - https://github.com/Preetwinder/frontera/blob/master/frontera/settings/default_settings.py#L21

sibiryakov: I think we should be using a single option for all backends. All these options adjust the same value but in different backends.
``STATE_CACHE_SIZE``
``HBASE_STATE_CACHE_SIZE_LIMIT``
``STATE_CACHE_SIZE_LIMIT``

In SQLA the STATE_CACHE_SIZE should be used instead of _LIMIT.

sibiryakov: I propose to fix that in a separate PR.

Issue #129, opened 2016-03-30
More efficient state cache in strategy workers


sibiryakov:
At the moment extracted links from spiders are distributed to SWs using fingerprint from crawled page. That causes same links to consume memory and HBase(or other DB) bandwidth on all SW instances. This can be improved by using additional message type in spider log, with key of extracted link fingerprint. Therefore links will be distributed uniformly to SWs and no duplication will take place.

This improvement requires changing spider log code in Frontera spider backend, protocols and SW.

The drawbacks are more messages in message queue systems.

PR #178, opened 2016-07-27
Extracted links to SW instances assignment fix

sibiryakov:
The goal is to get fixed guarantee that no matter from where link is extracted it will land to the same strategy worker. This is achieved by partitioning the ``spider_log`` and using hostname as a key.

Fixes https://github.com/scrapinghub/frontera/issues/129


sibiryakov:
Summarizing the changes: a single ``page_crawled`` event was divided on two - ``page_crawled`` and ``links_extracted``. Links extracted are divided per host and then sent by wire to many SW instances, depending on where particular host is assigned. This is a few lines in ``MessageBusBackend``.

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -112,10 +118,13 @@ def _request_from_object(self, obj):
 
     def decode(self, message):
         message = dict_to_bytes(super(Decoder, self).decode(message))
+        if message['type'] == 'links_extracted':
```
Preetwinder: Due to the conversion to bytes in the line above the values and keys will be bytes in message.

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -87,7 +90,10 @@ def decode(self, buffer):
         obj = unpackb(buffer)
         if obj[0] == b'pc':
             return ('page_crawled',
-                    self._response_from_object(obj[1]),
+                    self._response_from_object(obj[1]))
+        if obj[0] == 'le':
```
Preetwinder: I think `'le'` should be bytes here since unpackb return bytes.

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -78,3 +90,20 @@ def queue(self):
...
     def states(self):
         return None
+
+
+def aggregate_per_host(requests):
+    per_host = dict()
+    for link in requests:
+        if b'fingerprint' not in link.meta[b'domain']:
+            continue
+        host_fprint = link.meta[b'domain'][b'fingerprint']
```
Preetwinder: This is just a minor point, but maybe we could use the function defined below here. for e.g.

    for link in requests:
        host_fprint = get_host_fprint(link)
        if host_fprint is None:
            continue
        ...
The only issue I can think of with this code would be that it assigns a variable for all cases.

frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -112,8 +112,11 @@ def process_spider_output(self, response, result, spider):
                 links.append(element)
             else:
                 yield element
-        self.frontier.page_crawled(response=response,
-                                   links=links)
+        frontier_request = response.meta['frontier_request']
+        self.frontier.page_crawled(response)  # removed frontier part from .meta
+        # putting it back, to persist .meta from original request
+        response.meta['frontier_request'] = frontier_request
+        self.frontier.links_extracted(response.request, links)
```
Preetwinder: Shouldn't the keys here be bytes?

frontera/contrib/backends/remote/messagebus.py
sibiryakov: I think we should rely on DomainMiddleware always. In fact it should be hardcoded, because without it too many things will break.

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -40,13 +41,24 @@ def frontier_stop(self):
         self.spider_log_producer.flush()
 
     def add_seeds(self, seeds):
-        self.spider_log_producer.send(seeds[0].meta[b'fingerprint'], self._encoder.encode_add_seeds(seeds))
+        per_host = aggregate_per_host(seeds)
+        for host_fprint, host_links in per_host.iteritems():
```
Preetwinder: This should be `six.iteritems`.

PR #182, opened 2016-08-03
Tests for db and strategy workers

Preetwinder:
These tests will fail for PY3 right now since the workers haven't been yet ported.

### DIFF
--- a/tests/test_worker_db.py
+++ b/tests/test_worker_db.py
@@ -1,0 +1,64 @@
+from frontera.core.models import Request, Response
+from frontera.worker.db import DBWorker
+from frontera.settings import Settings
+from frontera.core.components import States
+
+
+r1 = Request('http://www.example.com/', meta={'fingerprint': '1', 'state': States.DEFAULT, 'jid': 0})
+r2 = Request('http://www.scrapy.org/', meta={'fingerprint': '2', 'state': States.DEFAULT, 'jid': 0})
+r3 = Request('https://www.dmoz.org', meta={'fingerprint': '3', 'state': States.DEFAULT, 'jid': 0})
+
+
+class TestDBWorker(object):
+
+    def dbw_setup(self, distributed=False):
+        settings = Settings()
+        settings.MESSAGE_BUS = 'tests.mocks.message_bus.FakeMessageBus'
+        if distributed:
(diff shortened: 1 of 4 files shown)


tests/test_worker_db.py
```diff
@@ -0,0 +1,64 @@
+from frontera.core.models import Request, Response
+from frontera.worker.db import DBWorker
+from frontera.settings import Settings
+from frontera.core.components import States
+
```
sibiryakov: There is a ``available_partitions`` and ``mark_ready``/``mark_busy`` mechanisms in message bus https://github.com/scrapinghub/frontera/blob/master/frontera/core/messagebus.py#L137. They are used mainly in DBW, and it would be nice to test them.

sibiryakov:
Overall, looks good!

tests/test_worker_db.py
```diff
@@ -0,0 +1,82 @@
...
+        assert set(dbw.spider_feed_producer.messages) == \
+            set([dbw._encoder.encode_request(r) for r in [r1, r2, r3]])
+
+    def test_offset(self):
+        dbw = self.dbw_setup(True)
+        msg = dbw._encoder.encode_offset(2, 50)
+        dbw.spider_log_consumer.put_messages([msg])
+        dbw.spider_feed_producer.offset = 100
+        dbw.consume_incoming()
+        assert 2 in dbw.spider_feed.ready_partitions
```
sibiryakov: ready_partitions isn't a part of public API, available_partitions() is.

tests/test_worker_strategy.py
```diff
@@ -0,0 +1,57 @@
...
+                for r in [r1, r3, r4]])
+
+    def test_page_crawled(self):
+        sw = self.sw_setup()
+        r1.meta['jid'] = 1
+        resp = Response(r1.url, request=r1)
+        msg = sw._encoder.encode_page_crawled(resp, [r2, r3, r4])
+        sw.consumer.put_messages([msg])
+        sw.work()
+        assert sw.scoring_log_producer.messages == []
```
sibiryakov: This is a test for skipping requests with other than SW job_ids? If so, let's put a comment saying about that. It's not obvious.

PR #184, opened 2016-08-03
Remove canonicalize_url

Preetwinder:


requirements.txt
```diff
@@ -1,2 +1,2 @@
 six>=1.8.0
-w3lib>=1.10.0
\ No newline at end of file
+w3lib>=1.15.0
```
sibiryakov: Also in setup.py!

Preetwinder: Oh yes. Silly me!

PR #185, opened 2016-08-04
Transfering of whole request over scoring log

sibiryakov:


frontera/core/codec.py
```diff
@@ -76,11 +76,11 @@ def encode_request(self, request):
...
     @abstractmethod
-    def encode_update_score(self, fingerprint, score, url, schedule):
+    def encode_update_score(self, request, score, schedule):
         """
         Encodes update_score messages for scoring log stream.
 
-        :param str fingerprint: fingerprint in hex form
+        :param object request: Frontera Request object
         :param float score: score
         :param str url: A document url
```
Preetwinder: Shouldn't this parameter be removed?

sibiryakov: Fingerprint is in request, so it doesn't make sense to transfer it.

Preetwinder: I am referring to the url parameter. The parameter was removed but it's documentation wasn't.

PR #186, opened 2016-08-04
HBase revisiting

sibiryakov:
Adding timestamp to the HBaseQueue and using current time to select items from the past.

### DIFF
--- a/frontera/contrib/backends/hbase.py
+++ b/frontera/contrib/backends/hbase.py
@@ -87,16 +87,20 @@ class HBaseQueue(Queue):
         pass
 
     def schedule(self, batch):
-        to_schedule = []
+        to_schedule = dict()
+        now = int(time())
         for fprint, score, request, schedule in batch:
             if schedule:
-                if 'domain' not in request.meta:
+                if 'domain' not in request.meta:    # TODO: this have to be done always by DomainMiddleware,
+                    # so I propose to require DomainMiddleware by HBaseBackend and remove that code
                     _, hostname, _, _, _, _ = parse_domain_from_url_fast(request.url)
                     if not hostname:
                         self.logger.error("Can't get hostname for URL %s, fingerprint %s", request.url, fprint)
                     request.meta['domain'] = {'name': hostname}
-                to_schedule.append((score, fprint, request.meta['domain'], request.url))
-        self._schedule(to_schedule, timestamp=int(time()))
+                dt = timegm(request.meta['crawl_at'].timetuple()) if 'crawl_at' in request.meta else now
+                to_schedule.setdefault(dt, []).append((request, score))
+        for timestamp, batch in to_schedule.iteritems():
+            self._schedule(batch, timestamp)
 
     def _schedule(self, batch, timestamp):
         """
@@ -114,7 +118,7 @@ class HBaseQueue(Queue):
           [0.99-1.00]
         random_str - the time when links was scheduled for retrieval, microsecs
 
-        :param batch: list of tuples(score, fingerprint, domain, url)
+        :param batch: iterable of Request objects
         :return:
         """
         def get_interval(score, resolution):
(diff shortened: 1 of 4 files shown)


frontera/contrib/backends/hbase.py
```diff
@@ -87,21 +87,25 @@ def frontier_stop(self):
...
+                    # so I propose to require DomainMiddleware by HBaseBackend and remove that code
                     _, hostname, _, _, _, _ = parse_domain_from_url_fast(request.url)
                     if not hostname:
                         self.logger.error("Can't get hostname for URL %s, fingerprint %s", request.url, fprint)
                     request.meta['domain'] = {'name': hostname}
-                to_schedule.append((score, fprint, request.meta['domain'], request.url))
-        self._schedule(to_schedule)
+                timestamp = request.meta['crawl_at'] if 'crawl_at' in request.meta else now
+                to_schedule.setdefault(timestamp, []).append((request, score))
+        for timestamp, batch in to_schedule.iteritems():
```
Preetwinder: I think that `six.iteritems` should be used here.

sibiryakov: Yep, I agree.

frontera/contrib/backends/hbase.py
```diff
@@ -22,6 +22,7 @@
 import six
 from six.moves import map
 from six.moves import range
+from w3lib.util import to_native_str
 
 
 _pack_functions = {
```
Preetwinder: I think that _pack_functions for URL and domain_fingerprint should be `to_bytes`. `str` returns native strings. I think we want happybase to receive bytes.

frontera/contrib/backends/hbase.py
```diff
@@ -133,27 +134,27 @@ def get_interval(score, resolution):
...
+                host_crc32 = get_crc32(domain[b'name'])
             elif type(domain) == int:
                 partition_id = self.partitioner.partition_by_hash(domain, self.partitions)
                 host_crc32 = domain
             else:
                 raise TypeError("domain of unknown type.")
             item = (unhexlify(fingerprint), host_crc32, request.url, score)
             score = 1 - score  # because of lexicographical sort in HBase
-            rk = "%d_%s_%d" % (partition_id, "%0.2f_%0.2f" % get_interval(score, 0.01), random_str)
+            rk = b"%d_%s_%d" % (partition_id, b"%0.2f_%0.2f" % get_interval(score, 0.01), random_str)
```
Preetwinder: This won't work for Python 3 versions less than 3.5. http://legacy.python.org/dev/peps/pep-0461/
I would suggest forming a normal string and then applying `to_bytes`. This needs to be changed in other parts of the PR as well.

frontera/contrib/backends/hbase.py
```diff
@@ -189,9 +190,9 @@ def get_next_requests(self, max_n_requests, partition_id, **kwargs):
         limit = min_requests
         tries = 0
         count = 0
-        prefix = '%d_' % partition_id
+        prefix = b'%d_' % partition_id
```
Preetwinder: If we use `to_bytes` for converting prefix and filter(as I suggested above) we end up with filter = `PrefixFilter b'0_'` thus we get an invalid prefix error. To avoid this we can either not convert prefix, or not convert both prefix and filter. This won't be a problem since happybase accepts native_string as argument for filter.

Preetwinder:
The reason for the current failure seems to be that the `start-hbase.sh`(in the hbase-docker repo) runs non-PY3 compliant code. As can be seen [here](https://github.com/dajobe/hbase-docker/blob/master/start-hbase.sh) the print statements lack parenthesis. I can either submit a PR there or we can use a [forked version](https://github.com/Preetwinder/hbase-docker).

sibiryakov:
@Preetwinder the project looks abandoned, let's fork it under scrapinghub, and then fix this fork. I'll arrange the public fork.

sibiryakov:
@Preetwinder please submit a PR here https://github.com/scrapinghub/hbase-docker and let's setup everything to make docker images pulling from using this repo.

PR #188, opened 2016-08-08
PY3 changes for codecs

Preetwinder:


frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -65,33 +68,34 @@ def __init__(self, request_model, response_model, *a, **kw):
...
-        return self._request_model(url=obj[0],
+        return self._request_model(url=to_native_str(obj[0]),
                                    headers=obj[1],
                                    cookies=obj[2],
                                    meta=obj[3])
 
     def decode(self, buffer):
         obj = unpackb(buffer)
-        if obj[0] == 'pc':
+        if obj[0] == b'pc':
```
sibiryakov: I think, usage of bytes type in encoders for message type will improve readability and redundancy. Because currently you rely on fact that msgpack after unpacking returns bytes (even if original string was native), that could change in the future, and it's non-obvious.

Preetwinder: I agree.

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -91,48 +93,49 @@ def __init__(self, request_model, response_model, *a, **kw):
...
     def decode_request(self, message):
-        obj = super(Decoder, self).decode(message)
-        return self._request_model(url=obj['url'],
-                                   method=obj['method'],
-                                   headers=obj['headers'],
-                                   cookies=obj['cookies'],
-                                   meta=obj['meta'])
+        obj = dict_to_bytes(super(Decoder, self).decode(message))
+        return self._request_model(url=to_native_str(obj[b'url']),
+                                   method=to_native_str(obj[b'method']),
```
sibiryakov: method here is converted to native string and above in ``_request_from_object() `` it isn't. I think there is a mistake here.

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -65,33 +68,34 @@ def __init__(self, request_model, response_model, *a, **kw):
         self._response_model = response_model
 
     def _response_from_object(self, obj):
-        return self._response_model(url=obj[0],
+        url = to_native_str(obj[0])
```
sibiryakov: you don't need a local variable here, probably.

Preetwinder: Do you think it is better to call `to_native_str` twice? Once for response_model and then for request_model inside it.

frontera/contrib/backends/remote/codecs/json.py
Preetwinder: I agree. I missed it.

frontera/contrib/backends/remote/codecs/msgpack.py
sibiryakov: I missed the second call, with locals it's better.

PR #189, opened 2016-08-09
Allow passing py.test args with tox

lopuhin:
This allows to run only specified tests under tox:

    tox -e py27,py35 -- tests/test_utils_misc.py::TestGetCRC32

When nothing is specified, "tests" is passed as default value, as before. I saw that in scrapy: https://github.com/scrapy/scrapy/blob/1075587dbd15e5ccb9a83c4ca14086c1e135fe12/tox.ini#L22. During development, this simplifies testing code that must work in different python versions, so you can check new tests or fixes with one command.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,7 +13,7 @@ deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/requirements/tests.txt
 commands =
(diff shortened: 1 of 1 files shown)


sibiryakov:
I think, that looks fine!
LGTM
@Preetwinder

Preetwinder:
+1. Looks Good!

PR #192, opened 2016-08-10
Py3 workers changes

Preetwinder:
Mostly fixing meta.

### DIFF
--- a/frontera/contrib/backends/remote/messagebus.py
+++ b/frontera/contrib/backends/remote/messagebus.py
@@ -38,13 +38,13 @@ class MessageBusBackend(Backend):
         self.spider_log_producer.flush()
 
     def add_seeds(self, seeds):
-        self.spider_log_producer.send(seeds[0].meta['fingerprint'], self._encoder.encode_add_seeds(seeds))
+        self.spider_log_producer.send(seeds[0].meta[b'fingerprint'], self._encoder.encode_add_seeds(seeds))
 
     def page_crawled(self, response, links):
-        self.spider_log_producer.send(response.meta['fingerprint'], self._encoder.encode_page_crawled(response, links))
+        self.spider_log_producer.send(response.meta[b'fingerprint'], self._encoder.encode_page_crawled(response, links))
 
     def request_error(self, page, error):
-        self.spider_log_producer.send(page.meta['fingerprint'], self._encoder.encode_request_error(page, error))
+        self.spider_log_producer.send(page.meta[b'fingerprint'], self._encoder.encode_request_error(page, error))
 
     def _get_next_requests(self, max_n_requests, **kwargs):
         requests = []
@@ -55,7 +55,7 @@ class MessageBusBackend(Backend):
                 self._logger.warning("Could not decode message: {0}, error {1}".format(encoded, str(exc)))
             else:
                 requests.append(request)
-        self.spider_log_producer.send('0123456789abcdef0123456789abcdef012345678',
+        self.spider_log_producer.send(b'0123456789abcdef0123456789abcdef012345678',
                                       self._encoder.encode_offset(self.partition_id, self.consumer.get_offset()))
         return requests
 
--- a/frontera/contrib/backends/sqlalchemy/components.py
+++ b/frontera/contrib/backends/sqlalchemy/components.py
@@ -14,7 +14,7 @@ from frontera.utils.misc import get_crc32, chunks
 from frontera.utils.url import parse_domain_from_url_fast
 import six
 from six.moves import range
-from w3lib.util import to_native_str
+from w3lib.util import to_native_str, to_bytes
 
 
 def retry_and_rollback(func):
--- a/frontera/contrib/backends/sqlalchemy/components.py
+++ b/frontera/contrib/backends/sqlalchemy/components.py
@@ -75,7 +75,7 @@ class Metadata(BaseMetadata):
         db_page.fetched_at = datetime.utcnow()
         if isinstance(obj, Response):
             db_page.headers = obj.request.headers
-            db_page.method = obj.request.method
+            db_page.method = to_native_str(obj.request.method)
             db_page.cookies = obj.request.cookies
             db_page.status_code = obj.status_code
         return db_page
@@ -90,11 +90,11 @@ class Metadata(BaseMetadata):
 
         if isinstance(obj, Request):
             db_page.headers = obj.headers
-            db_page.method = obj.method
+            db_page.method = to_native_str(obj.method)
             db_page.cookies = obj.cookies
         elif isinstance(obj, Response):
             db_page.headers = obj.request.headers
-            db_page.method = obj.request.method
+            db_page.method = to_native_str(obj.request.method)
             db_page.cookies = obj.request.cookies
             db_page.status_code = obj.status_code
         return db_page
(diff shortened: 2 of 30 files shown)


frontera/contrib/scrapy/converters.py
```diff
@@ -50,20 +51,20 @@ def to_frontier(self, scrapy_request):
...
-        meta = frontier_request.meta.get('scrapy_meta', {})
-        meta['frontier_request'] = frontier_request
+        meta = frontier_request.meta.get(b'scrapy_meta', {})
+        meta[b'frontier_request'] = frontier_request
         return ScrapyRequest(url=frontier_request.url,
                              callback=cb,
                              errback=eb,
                              body=body,
-                             method=frontier_request.method,
+                             method=to_native_str(frontier_request.method),
```
Preetwinder: Scrapy uses native strings for method. If we pass bytes, it deforms the string.

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -173,10 +173,10 @@ def get_next_requests(self, max_n_requests, partition_id, **kwargs):
         try:
             for item in self._order_by(self.session.query(self.queue_model).filter_by(partition_id=partition_id)).\
                     limit(max_n_requests):
-                method = item.method or 'GET'
+                method = item.method or b'GET'
```
sibiryakov: What type item.method will get? In case of sqlite and postgresql? I thought we agreed on native_str before SQLA and bytes anywhere else in Frontera.

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -173,10 +173,10 @@ def get_next_requests(self, max_n_requests, partition_id, **kwargs):
...
                 r = Request(item.url, method=method, meta=item.meta, headers=item.headers, cookies=item.cookies)
-                r.meta['fingerprint'] = to_native_str(item.fingerprint)
-                r.meta['score'] = item.score
+                r.meta[b'fingerprint'] = to_native_str(item.fingerprint)
```
sibiryakov: I thought we agreed that both meta values and keys will be encoded in bytes?

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -266,7 +266,7 @@ def get_next_requests(self, max_n_requests, partition_id, **kwargs):
         results = []
         for items in six.itervalues(queue):
             for item in items:
-                method = item.method or 'GET'
+                method = item.method or b'GET'
```
sibiryakov: the same about item.method here.

frontera/contrib/backends/sqlalchemy/components.py
Preetwinder: This is a mistake. I'll remove the conversion.

frontera/contrib/middlewares/domain.py
```diff
@@ -87,10 +87,10 @@ def request_error(self, request, error):
         return self._add_domain(request)
 
     def _add_domain(self, obj):
-        obj.meta['domain'] = self.parse_domain_info(obj.url, self.manager.test_mode)
```
sibiryakov: What is the type of results from parse_domain_info?

frontera/contrib/backends/sqlalchemy/components.py
Preetwinder: The type here doesn't matter because in the next line we initialize the Request object with method and it is converted to bytes in the Request model.

sibiryakov: ok

frontera/contrib/backends/sqlalchemy/components.py
Preetwinder: The same here. method is solely used to initialize Request. And we convert the method argument to bytes in the Request model.

frontera/contrib/middlewares/domain.py
Preetwinder: It returns a dictionary with netloc, name, scheme, etc. The keys are bytes and the values are native strings. Do you think it would be better to convert the values too?

sibiryakov: If we plan to declare that everything in meta has type bytes, then yes.

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -51,20 +51,20 @@ def frontier_stop(self):
     def add_seeds(self, seeds):
         for seed in seeds:
             o = self._create_page(seed)
-            self.cache[o.fingerprint] = self.session.merge(o)
+            self.cache[to_bytes(o.fingerprint)] = self.session.merge(o)
```
Preetwinder: I have used bytes as keys for the cache in both metadata and state. The reason is to reduce the number of conversions needed in other parts of the code.

sibiryakov: That's right.

tests/contrib/backends/hbase/test_hbase.py
```diff
@@ -58,7 +58,7 @@ def test_queue_with_delay(self):
         queue.schedule(batch)
         assert queue.get_next_requests(10, 0, min_requests=3, min_hosts=1,
                    max_requests_per_host=10) == []
-        sleep(1.0)
+        sleep(1.5)
```
sibiryakov: :) I believe that would be enough :)

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -38,13 +38,13 @@ def frontier_stop(self):
         self.spider_log_producer.flush()
```
sibiryakov: Is this class tested somewhere? It's minor, but I don't see any code testing it.

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -123,18 +123,18 @@ def frontier_stop(self):
 
     @retry_and_rollback
     def fetch(self, fingerprints):
-        to_fetch = [f for f in fingerprints if f not in self._cache]
+        to_fetch = [to_native_str(f) for f in fingerprints if f not in self._cache]
```
sibiryakov: Here you search for native strings in `self._cache`

frontera/contrib/backends/sqlalchemy/components.py
```diff
@@ -123,18 +123,18 @@ def frontier_stop(self):
...
         self.logger.debug("cache size %s", len(self._cache))
         self.logger.debug("to fetch %d from %d", len(to_fetch), len(fingerprints))
 
         for chunk in chunks(to_fetch, 128):
             for state in self.session.query(self.model).filter(self.model.fingerprint.in_(chunk)):
-                self._cache[str(state.fingerprint)] = state.state
+                self._cache[to_bytes(state.fingerprint)] = state.state
```
sibiryakov: And here you store bytes to `self._cache`.

please also check if it fits the code in parent class https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/memory/__init__.py#L122
I see there unicode requested from meta for PY3.

frontera/contrib/backends/sqlalchemy/components.py
Preetwinder: Since `f` is iterated over `fingerprints`(which contains fingerprints as bytes), the search is carried out using f as bytes. The resulting list will have these fingerprints as native_str. `to_fetch` is then used to query the DB where we want them to be as native strings.

frontera/contrib/backends/sqlalchemy/components.py
Preetwinder: We want to use bytes for cache. MemoryState doesn't perform any conversions, it uses fingerprints as bytes. So I don't think there should be a compatibility issue.

frontera/utils/tester.py
```diff
@@ -74,10 +74,10 @@ def _run_iteration(self):
             else:
                 self.frontier.request_error(request=page_to_crawl,
                                             error=crawled_page.status)
-            assert page_to_crawl.meta['this_param'] == 'should be passed over'
-            assert page_to_crawl.headers['X-Important-Header'] == 'Frontera'
-            assert page_to_crawl.method == 'POST'
-            assert page_to_crawl.cookies['currency'] == 'USD'
+            assert page_to_crawl.meta[b'this_param'] == 'should be passed over'
```
sibiryakov: 'should be passed over' needs to be bytes too

PR #193, opened 2016-08-10
Codec option

sibiryakov:
option and docs update


Preetwinder:
Looks good. I am guessing we want to only use msgpack for request persistence in HBase queue?

sibiryakov:
I think this is improbable, because HBase is used when user is interested in large scale deployment and performance, msgpack provides both. If someone would need JSON, or even some custom format, I believe this is very custom and we shouldn't care about this.

PR #196, opened 2016-08-15
Py3 messagebus

Preetwinder:


frontera/contrib/messagebus/zeromq/broker.py
```diff
@@ -107,23 +108,23 @@ def handle_db_out_recv(self, msg):
 
     def handle_db_in_recv(self, msg):
         self.stats['db_in_recvd'] += 1
-        if msg[0][0] in ['\x01', '\x00']:
+        if to_native_str(msg[0])[0] in ['\x01', '\x00']:
```
sibiryakov: what is the type of `msg[0]`? is it possible to perform such comparision using bytes always?

sibiryakov:
Awesome!!! All checks are passed !!!

frontera/contrib/messagebus/zeromq/broker.py
Preetwinder: msg[0] is bytes always. The comparison works in PY2, but in PY3 the characters of the bytestring are interpreted as integers so `1 in ['\x01', '\x00']` fails. Operations such as these require unicode in PY3.

sibiryakov: maybe the cheaper would be to compare character code of first character?

Preetwinder: We can use - `if b'\x01' in msg[0] or b'\x00' in msg[0]:`. The other option would be for PY2 to use `msg[0][0]` as is(because it will be a byte) and for PY3 convert `msg[0][0]` to character code(from int).

sibiryakov: I vote for a first option.

PR #205, opened 2016-09-13
Added tests for message_bus_backend

voith:


voith:
ping @sibiryakov

tests/test_message_bus_backend.py
```diff
@@ -0,0 +1,78 @@
...
+        mbb.add_seeds([r1, r2, r3])
+        seeds = [mbb._decoder.decode(m)[1][0] for m in mbb.spider_log_producer.messages]
+        self.assertEqual(set([seed.url for seed in seeds]), set([r1.url, r2.url, r3.url]))
+
+    def test_page_crawled(self):
+        mbb = self.mbb_setup()
+        resp = Response(r1.url, body='body', request=r1)
+        mbb.page_crawled(resp)
+        page = mbb._decoder.decode(mbb.spider_log_producer.messages[0])[1]
+        self.assertEqual(page.request.url, resp.request.url)
```
sibiryakov: Considering that you enabled STORE_CONTENT, I think there's body check missing.

tests/test_message_bus_backend.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_links_extracted(self):
+        mbb = self.mbb_setup()
+        mbb.links_extracted(r1, [r2, r3])
+        links = [mbb._decoder.decode(m)[2][0] for m in mbb.spider_log_producer.messages]
+        self.assertEqual(set([link.url for link in links]), set([r2.url, r3.url]))
```
sibiryakov: The request check (r1) is missing.

tests/test_message_bus_backend.py
```diff
@@ -0,0 +1,78 @@
...
+        mbb = self.mbb_setup()
+        mbb.request_error(r1, 'error')
+        _, error_request, error_message = mbb._decoder.decode(mbb.spider_log_producer.messages[0])
+        self.assertEqual((error_request.url, error_message), (r1.url, 'error'))
+
+    def test_get_next_requests(self):
+        mbb = self.mbb_setup()
+        encoded_requests = [mbb._encoder.encode_request(r) for r in [r1, r2, r3]]
+        mbb.consumer.put_messages(encoded_requests)
+        mbb.consumer._set_offset(0)
```
sibiryakov: MessageBusBackend also generated offset-type messages, what do you think of checking them too?

voith:
@sibiryakov Thanks for reviewing my PR. I agree with all your suggestions. I will get back to this soon.

tests/test_message_bus_backend.py
voith: @sibiryakov I did not understand what `offset-type` messages mean? I thought that `type(offset)` is always `int`. Do you mean to check `decoded_message_type == 'offset'`? Or can you direct me to the code that you're referring to, so that I can it understand better.

sibiryakov: When get_next_requests() is called the offset message is generated and send by means of message bus. I suggest to add assert checking this message too. Currently your code is checking only returned values.

sibiryakov: https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/remote/messagebus.py#L72

tests/test_message_bus_backend.py
voith: applied suggestion

tests/test_message_bus_backend.py
voith: applied suggestion

tests/test_message_bus_backend.py
voith: Thanks for making this easy to understand. I've applied this suggestion

voith:
@sibiryakov I've made the suggested changes. Please let me know if you have any other suggestions.

sibiryakov:
thank you 🍺

PR #212, opened 2016-10-05
[WIP] Fix #211

### LINKED ISSUE #211: frontera converts keys in scrapy's response meta to bytes
while running crawler in https://github.com/sibiryakov/frontera-google I get the following error.

```
2016-10-04 23:38:19 [scrapy] ERROR: Spider error processing <GET http://zpravy.idnes.cz/madarsko-zmena-ustavy-zakaz-kvot-migrace-f1c-/zahranicni.aspx?c=A161004_175512_zahranicni_mlb> (referer: None)
  File "/home/voith/Projects/frontera-google/env-3.5/lib/python3.5/site-packages/twisted/internet/defer.py", line 587, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/voith/Projects/frontera-google/env-3.5/lib/python3.5/site-packages/scrapy/spiders/crawl.py", line 68, in _response_downloaded
    rule = self._rules[response.meta['rule']]
KeyError: 'rule'
```

I ran the crawler without using frontera(only using scrapy) and it ran fine. That gave me a hint that frontera must be causing the error.
from the little debugging that I did, I figured out the following:

```
> /home/voith/Projects/frontera-google/env-3.5/lib/python3.5/site-packages/scrapy/spiders/crawl.py(68)_response_downloaded()
-> rule = self._rules[response.meta['rule']]
(Pdb) response.meta
{'download_timeout': 180.0, b'frontier_request': <Request at 0x7f1b39a9cc50 https://www.alza.cz/nejobchod meta={b'scrapy_callback': b'_response_downloaded', b'score': 0.16666666666666666, b'scrapy_errback': None, b'scrapy_meta': {b'rule': 0, b'frontier_request': <Request at 0x7f1b39a9cc50 https://www.alza.cz/nejobchod meta={...} body=... cookies={}, headers={}>, b'link_text': b'\r\n\r\n'}, b'domain': {b'netloc': b'www.alza.cz', b'tld': b'', b'subdomain': b'', b'fingerprint': b'4273593b6eda0469d3db862aa0fda848d7a89974', b'name': b'www.alza.cz', b'scheme': b'https', b'sld': b''}, b'fingerprint': b'7ee99dd119df89a7a103e1001b97f1255ea6e466', b'state': 1, b'origin_is_frontier': True, b'jid': 0} body=... cookies={}, headers={}>, b'rule': 0, 'download_latency': 0.33022451400756836, b'link_text': b'\r\n\r\n', 'download_slot': 'www.alza.cz'}

(Pdb) response.meta['rule']
*** KeyError: 'rule'

(Pdb) response.meta[b'rule']
0

```

Since the keys have been converted into bytes, they cannot be accessed by just using the key name.

I have yet to figure which part of the code is exactly causing this. If I figure it out, I'll definitely submit a fix for this. But I'm logging this for now so that it can be tracked.

chuckus:
This is definitely a work-in-progress but just want to check if I am on the right path? I am running stock https://github.com/sibiryakov/frontera-google and not getting any KeyErrors, but am receiving some new errors but I think are unrelated (no `'redirect_*'` info in `meta`).
#211 Convert scrapy meta back to unicode when decoding msgpack encoding messages

### DIFF
--- a/frontera/contrib/scrapy/converters.py
+++ b/frontera/contrib/scrapy/converters.py
@@ -7,6 +7,8 @@ from frontera.core.models import Response as FrontierResponse
 from frontera.utils.converters import BaseRequestConverter, BaseResponseConverter
 from w3lib.util import to_bytes, to_native_str
 
+from frontera.utils.misc import dict_to_unicode, dict_to_bytes
+
 
 class RequestConverter(BaseRequestConverter):
     """Converts between frontera and Scrapy request objects"""
@@ -26,7 +28,7 @@ class RequestConverter(BaseRequestConverter):
         if callable(eb):
             eb = _find_method(self.spider, eb)
 
-        scrapy_meta = scrapy_request.meta
+        scrapy_meta = dict_to_bytes(scrapy_request.meta, keys_only=True)
         meta = {}
         if b'frontier_request' in scrapy_meta:
             request = scrapy_meta[b'frontier_request']
@@ -58,7 +60,8 @@ class RequestConverter(BaseRequestConverter):
         if eb and self.spider:
             eb = _get_method(self.spider, eb)
         body = frontier_request.body
-        meta = frontier_request.meta.get(b'scrapy_meta', {})
+        meta = dict_to_unicode(frontier_request.meta.get(b'scrapy_meta', {}),
+                               keys_only=True)
         meta[b'frontier_request'] = frontier_request
         return ScrapyRequest(url=frontier_request.url,
                              callback=cb,
--- a/tests/test_scrapy.py
+++ b/tests/test_scrapy.py
@@ -43,7 +43,7 @@ def test_request_response_converters():
     assert b'frontier_request' not in frontier_request.meta[b'scrapy_meta']
 
     request_converted = rc.from_frontier(frontier_request)
-    assert request_converted.meta[b'test_param'] == b'test_value'
+    assert request_converted.meta['test_param'] == b'test_value'
     assert request_converted.body == to_bytes(REQUEST_BODY)
     assert request_converted.url == url
     assert request_converted.method == 'GET'
@@ -60,14 +60,14 @@ def test_request_response_converters():
 
     frontier_response = rsc.to_frontier(response)
     assert frontier_response.body == RESPONSE_BODY
-    assert frontier_response.meta[b'scrapy_meta'][b'test_param'] == b'test_value'
+    assert frontier_response.meta[b'scrapy_meta']['test_param'] == b'test_value'
     assert frontier_response.meta[b'scrapy_meta'][b'middleware_stuff'] == b'appeared'
     assert frontier_response.status_code == 200
     assert b'frontier_request' not in frontier_response.meta[b'scrapy_meta']
 
     response_converted = rsc.from_frontier(frontier_response)
     assert response_converted.body == RESPONSE_BODY
-    assert response_converted.meta[b'test_param'] == b'test_value'
+    assert response_converted.meta['test_param'] == b'test_value'
     assert response_converted.url == url
     assert response_converted.status == 200
     assert response_converted.headers[b'TestHeader'] == b'Test value'
--- a/frontera/contrib/backends/remote/codecs/msgpack.py
+++ b/frontera/contrib/backends/remote/codecs/msgpack.py
@@ -6,6 +6,8 @@ from __future__ import absolute_import
 from msgpack import packb, unpackb
 
 from frontera.core.codec import BaseDecoder, BaseEncoder
+from frontera.utils.misc import dict_to_unicode
+
 import six
 from w3lib.util import to_native_str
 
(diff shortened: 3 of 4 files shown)


sibiryakov:
Hey @chuckus thanks for the PR! 
There is a two ways how string can be passed in Py3: native - (bytes in py2 and unicode in py3) and bytes (bytes in py2 and bytes in py3). Frontera requires all the content to be bytes (http://frontera.readthedocs.io/en/latest/topics/frontier-objects.html#frontera.core.models.Request.meta), because it's distributed and request objects need to be serialized without loss.

Scrapy is designed the other way: it expects code to be more pythonic, and therefore uses native strings.

If we start converting to native strings all the meta content from Frontera then we create another problem for those who stored data using bytes type and want's it back with the same type.

At the moment I have no idea what to do with this. The best _probably_ would be to preserve the string type in meta.

chuckus:
@sibiryakov, thanks for the explanation, I'll keep thinking of a more comprehensive solution and use `six` to maintain compatibility.

chuckus:
@sibiryakov, I think this new solution better as it does the conversion where it belongs. Please let me know what you think and anything you think I need to change/add. I successfully tested this again manually with your original frontera-google repo before my merged change to the spider.

chuckus:
Aware of the failure, will look into fixing it soon.

tests/test_scrapy.py
```diff
@@ -60,14 +60,14 @@ def test_request_response_converters():
 
     frontier_response = rsc.to_frontier(response)
     assert frontier_response.body == RESPONSE_BODY
-    assert frontier_response.meta[b'scrapy_meta'][b'test_param'] == b'test_value'
+    assert frontier_response.meta[b'scrapy_meta']['test_param'] == b'test_value'
```
chuckus: @sibiryakov what do you think of this particular assertion? should we have the entirety of `meta` as bytes when sending over to Frontera?

sibiryakov: @Preetwinder please have a look.

frontera/contrib/scrapy/converters.py
```diff
@@ -7,6 +7,8 @@
 from frontera.utils.converters import BaseRequestConverter, BaseResponseConverter
 from w3lib.util import to_bytes, to_native_str
 
+from frontera.utils.misc import dict_to_unicode, dict_to_bytes
+
 
 class RequestConverter(BaseRequestConverter):
```
voith: Please have a look at https://github.com/scrapinghub/frontera/issues/211#issuecomment-251931413
I am not very sure if this is the best place to resolve the issue. Although fixing this at `from_frontier` would solve the issue, my concern is that an encoder-decoder should not change the original form of a message on decoding.

frontera/contrib/scrapy/converters.py
```diff
@@ -26,7 +28,7 @@ def to_frontier(self, scrapy_request):
         if callable(eb):
             eb = _find_method(self.spider, eb)
 
-        scrapy_meta = scrapy_request.meta
+        scrapy_meta = dict_to_bytes(scrapy_request.meta, keys_only=True)
```
voith: Is this line really required? My thinking is that since this code is in `to_frontier`, scrapy's meta will be in the original form as scrapy generated it.

tests/test_scrapy.py
Preetwinder: Although in this case we can go either way, since frontera doesn't currently acess `scrapy_meta`, it might be better to follow the convention and have everything possible as bytes. So I would suggest a conversion to `bytes` for both keys and values.

frontera/contrib/scrapy/converters.py
Preetwinder: Although this won't make a difference to frontera since it currently doesn't access scrapy_meta, but we might want to keep the conversion of keys(and also values) to bytes to follow our convention of having bytes in meta.

frontera/contrib/scrapy/converters.py
Preetwinder: I personally think it would be better to solve this issue here for the following reasons.
1. The issue is specific to Scrapy, so it makes sense to solve it in the scrapy converter.
2. Making the conversion in encoder/decoder leads to an unnecessary overhead for other components which use them.

sibiryakov: I don't see how we could fix this for all possible Scrapy customizations (middlewares, spiders code). Scrapy perfectly allows user to store keys in values using any type and then assures user that he could rely on that storage, and there will be no changes later. 
I'm afraid the only way to fix that in all possible scenarios is to persist string type during serialization.

frontera/contrib/scrapy/converters.py
chuckus: I agree with @Preetwinder, correct me if I'm wrong @sibiryakov but over within frontera, you want to encode everything in bytes so that it Python version agnostic; important for a distributed system which may have Python 2 and Python 3 on different nodes. I'm going to go ahead and reimplement the above to deal with preserving all type information. The only thing I haven't decided upon is what happens when the meta is passed from Scrapy running Py2 to Scrapy running Py3 and vice versa. 

e.g. for `scrapy_meta`, Python 3 string (`unicode`) -> `to_frontier` -> (`bytes`) -> `from_frontier` -> Python 2 (`str` or `unicode`?)

sibiryakov: The answer to your question is in the code. Here it is written, that without ``use_bin_type=True`` the string content is always decoded to bytes (or str in Py2). https://pypi.python.org/pypi/msgpack-python
BTW @Preetwinder msgpack-python provides out of the box support of string type encoding. Probably we should check closer, and may be relax rules for meta content.

voith:
#211 has been closed. Can we close this PR now?

PR #216, opened 2016-10-06
fixes #215

### LINKED ISSUE #215: While using python 3.x HBASE_DROP_ALL_TABLES doesn't work
In python even after setting `HBASE_DROP_ALL_TABLES=True`, I get the following error.
```
AlreadyExists: AlreadyExists(message='table name already in use')
```
So I debugged this and figured out that hbase return all strings as bytes. When table name is set in setting (HBASE_METADATA_TABLE = 'metadata') it is passed to hbase as `str`. Hence the following line fails: https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/hbase.py#L77 and the tables are not dropped
```
self.table =  'metadata'
tables = tables = set(self.connection.tables())
print(tables) ---> {b'metadata', b'queue'}
'metadata' in {b'metadata', b'queue'} ---> False 

```

voith:
fixes #215

### DIFF
--- a/tests/contrib/backends/hbase/test_hbase.py
+++ b/tests/contrib/backends/hbase/test_hbase.py
@@ -87,3 +87,26 @@ class TestHBaseBackend(object):
         assert r4.meta[b'state'] == States.CRAWLED
         state.flush(True)
         assert state._state_cache == {}
+
+    def test_drop_all_tables_when_table_name_is_str(self):
+        connection = Connection(host='hbase-docker', port=9090)
+        for table in connection.tables():
+            connection.delete_table(table, True)
+        hbase_queue_table = 'queue'
+        queue_schema = {'f': {'max_versions': 1, 'block_cache_enabled': 1}}
+        hbase_metadata_table = 'metadata'
+        metadata_schema = {'m': {'max_versions': 1},
+                           's': {'max_versions': 1, 'block_cache_enabled': 1,
+                                 'bloom_filter_type': 'ROW', 'in_memory': True, },
+                           'c': {'max_versions': 1}
+                           }
+        connection.create_table(hbase_queue_table, queue_schema)
+        connection.create_table(hbase_metadata_table, metadata_schema)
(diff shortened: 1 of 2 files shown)


tests/contrib/backends/hbase/test_hbase.py
```diff
@@ -87,3 +87,26 @@ def test_state(self):
         assert r4.meta[b'state'] == States.CRAWLED
         state.flush(True)
         assert state._state_cache == {}
+
+    def test_drop_all_tables_when_table_name_is_str(self):
+        connection = Connection(host='hbase-docker', port=9090)
+        for table in connection.tables():
+            connection.delete_table(table, True)
+        hbase_queue_table = 'queue'
+        queue_schema = {'f': {'max_versions': 1, 'block_cache_enabled': 1}}
```
sibiryakov: I would reduce schemas here, they don't need to be that detailed.

tests/contrib/backends/hbase/test_hbase.py
```diff
@@ -87,3 +87,26 @@ def test_state(self):
...
+                           }
+        connection.create_table(hbase_queue_table, queue_schema)
+        connection.create_table(hbase_metadata_table, metadata_schema)
+        tables = connection.tables()
+        assert set(tables) == set([b'metadata', b'queue'])  # Failure of test itself
+
+        # the following two lines should fail with a table already exists exception
+        # if the table are not dropped successfully in the constructor
+        HBaseQueue(connection=connection, partitions=1, table_name=hbase_queue_table, drop=True)
+        HBaseMetadata(connection=connection, table_name=hbase_metadata_table, drop_all_tables=True,
```
sibiryakov: exception raise assertion is missing here.

tests/contrib/backends/hbase/test_hbase.py
voith: Done

PR #217, opened 2016-10-06
Alternative fix to #211 by forcing scrapy meta encoding to unicode at messagebus

chuckus:
Given the discussion in #211 and my original attempt to fix it in #212, I thought I'd make another PR so we could make a decision between the two and simply abandon the other. Thank you @voith for weighing in, I pretty much used your repl code verbatim for the test.

IMO, this isn't ideal as we are making an assumption here. The best solution would be to preserve `meta` encoding by generating and storing the `meta`'s _meta_ in the outgoing message, which should provide enough information for us to decode it properly at he other end, but that would be more work, do wanted everyone's opinion first. I'm happy to look into that if everyone agrees that is way forward.

### DIFF
--- a/frontera/contrib/backends/remote/codecs/msgpack.py
+++ b/frontera/contrib/backends/remote/codecs/msgpack.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 from msgpack import packb, unpackb
 
 from frontera.core.codec import BaseDecoder, BaseEncoder
+from frontera.utils.misc import dict_to_unicode
 import six
 from w3lib.util import to_native_str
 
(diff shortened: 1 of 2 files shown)


voith:
Thanks for the PR @chuckus. I too am not sure if this is the right way to go forward. Although I myself suggested this fix, this looks very hacky to me.

> The best solution would be to preserve meta encoding by generating and storing the meta's meta in the outgoing message, which should provide enough information for us to decode it properly at he other end

I would prefer this approach and If you could implement something like that than I would be more than happy to review it.  But wait till @sibiryakov gives you his opinion.

voith:
A few pointers after having a quick look through at your PR. Tests for codecs are written here: https://github.com/scrapinghub/frontera/blob/master/tests/test_codecs.py. So there's no need of creating a new file for it. Also there is another codec here: https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/backends/remote/codecs/json.py
But only make these changes once @sibiryakov gives you his opinion.

chuckus:
Thank you very much for your pointers @voith, will wait for @sibiryakov's opinion.

sibiryakov:
the answer is here https://github.com/scrapinghub/frontera/pull/212#issuecomment-251695665

voith:
@chuckus I have opened a fix for this over [here](#233). You could help me by reviewing the PR

voith:
#211 has been closed. Can we close this PR now?

PR #230, opened 2016-11-21
Correction in doc and added import in example

anujmax:


anujmax:
@sibiryakov I rebased as you suggested. Please review.

sibiryakov:
thanks!

PR #233, opened 2016-12-10
Fix codecs

### LINKED ISSUE #211: frontera converts keys in scrapy's response meta to bytes

voith:
Issue discussed here https://github.com/scrapinghub/frontera/issues/211#issuecomment-251931413
Todo List
- [X] Fix msgpack codec
- [x] Fix json codec
- [x] Integration test with Hbase backend(manually)

This PR fixes #211

Other things done in this besides the todo list:
- Added two methods `_convert` and `reconvert` in json codec. These are needed as JSONEncoder accepts strings only as unicode. Method `convert` converts objects recursively to unicode and saves their type.
- made the requirement of msgpack >=0.4 as only versions greater than 0.4 support the changes made in this PR.
- fixed a buggy test case in `test_message_bus_backend` which got exposed after fixing the codecs.

### DIFF
--- a/frontera/contrib/backends/remote/codecs/json.py
+++ b/frontera/contrib/backends/remote/codecs/json.py
@@ -3,18 +3,64 @@
 """
 from __future__ import absolute_import
 import json
+import six
 from base64 import b64decode, b64encode
 from frontera.core.codec import BaseDecoder, BaseEncoder
-from w3lib.util import to_unicode, to_native_str
-from frontera.utils.misc import dict_to_unicode, dict_to_bytes
+from w3lib.util import to_unicode, to_bytes
+
+
+def _convert_and_save_type(obj):
+    """
+    :param obj: dict object
+
+    The purpose of this method is to transform the given dict
+    into a form that would be able to serialize with JSONEncoder.
+    In order to implement this, this method converts all byte strings
+    inside a dict to unicode and saves their type for reverting to its
+    original state. The type and the value are stored as a tuple in the
+    following format: (original_type, converted value). All other objects
+    like dict, tuple, list are converted to the same format for the sake
+    of serialization and for the ease of reverting.
+    Refer `https://github.com/scrapinghub/frontera/pull/233#discussion_r97432868`
+    for the detailed explanation about the design.
+    """
+    if isinstance(obj, bytes):
+        return 'bytes', to_unicode(obj)
+    elif isinstance(obj, dict):
+        return 'dict', [(_convert_and_save_type(k), _convert_and_save_type(v)) for k, v in six.iteritems(obj)]
+    elif isinstance(obj, (list, tuple)):
+        return type(obj).__name__, [_convert_and_save_type(item) for item in obj]
+    return 'other', obj
+
+
+def _convert_from_saved_type(obj):
+    """
+    :param obj: object returned by `_convert_and_save_type`
+
+    Restores the original state of the object converted
+    earlier by `_convert_and_save_type`. This method considers every
+    first element of the nested tuple as the original type information and
+    the second value to be the converted value. It applies the original type
+    recursively on the object to retrieve the original form of the object.
+    """
+    assert len(obj) == 2
+    obj_type, obj_value = obj
+    if obj_type == 'bytes':
+        return to_bytes(obj_value)
+    elif obj_type == 'dict':
+        return dict([(_convert_from_saved_type(k), _convert_from_saved_type(v)) for k, v in obj_value])
+    elif obj_type in ['list', 'tuple']:
+        _type = list if obj_type == 'list' else tuple
+        return _type([_convert_from_saved_type(item) for item in obj_value])
+    return obj_value
 
 
 def _prepare_request_message(request):
-    return {'url': to_unicode(request.url),
-            'method': to_unicode(request.method),
-            'headers': dict_to_unicode(request.headers),
-            'cookies': dict_to_unicode(request.cookies),
-            'meta': dict_to_unicode(request.meta)}
+    return {'url': request.url,
+            'method': request.method,
+            'headers': request.headers,
+            'cookies': request.cookies,
+            'meta': request.meta}
 
 
 def _prepare_links_message(links):
@@ -22,10 +68,10 @@ def _prepare_links_message(links):
 
 
 def _prepare_response_message(response, send_body):
-    return {'url': to_unicode(response.url),
+    return {'url': response.url,
             'status_code': response.status_code,
-            'meta': dict_to_unicode(response.meta),
-            'body': to_unicode(b64encode(response.body)) if send_body else None}
+            'meta': response.meta,
+            'body': b64encode(response.body) if send_body else None}
 
 
 class CrawlFrontierJSONEncoder(json.JSONEncoder):
@@ -44,6 +90,10 @@ class Encoder(BaseEncoder, CrawlFrontierJSONEncoder):
     def __init__(self, request_model, *a, **kw):
         self.send_body = kw.pop('send_body', False)
         super(Encoder, self).__init__(request_model, *a, **kw)
+
+    def encode(self, obj):
+        encoded = _convert_and_save_type(obj)
+        return super(Encoder, self).encode(encoded)
 
     def encode_add_seeds(self, seeds):
         return self.encode({
@@ -101,52 +151,51 @@ class Decoder(json.JSONDecoder, BaseDecoder):
         super(Decoder, self).__init__(*a, **kw)
 
     def _response_from_object(self, obj):
-        url = to_native_str(obj[b'url'])
+        url = obj['url']
         request = self._request_model(url=url,
-                                      meta=obj[b'meta'])
+                                      meta=obj['meta'])
         return self._response_model(url=url,
-                                    status_code=obj[b'status_code'],
-                                    body=b64decode(obj[b'body']) if obj[b'body'] is not None else None,
+                                    status_code=obj['status_code'],
+                                    body=b64decode(obj['body']) if obj['body'] is not None else None,
                                     request=request)
 
     def _request_from_object(self, obj):
-        return self._request_model(url=to_native_str(obj[b'url']),
-                                   method=obj[b'method'],
-                                   headers=obj[b'headers'],
-                                   cookies=obj[b'cookies'],
-                                   meta=obj[b'meta'])
+        return self._request_model(url=obj['url'],
+                                   method=obj['method'],
+                                   headers=obj['headers'],
+                                   cookies=obj['cookies'],
+                                   meta=obj['meta'])
 
     def decode(self, message):
-        message = dict_to_bytes(super(Decoder, self).decode(message))
-        if message[b'type'] == b'links_extracted':
-            request = self._request_from_object(message[b'r'])
-            links = [self._request_from_object(link) for link in message[b'links']]
+        message = _convert_from_saved_type(super(Decoder, self).decode(message))
+        if message['type'] == 'links_extracted':
+            request = self._request_from_object(message['r'])
+            links = [self._request_from_object(link) for link in message['links']]
             return ('links_extracted', request, links)
-        if message[b'type'] == b'page_crawled':
-            response = self._response_from_object(message[b'r'])
+        if message['type'] == 'page_crawled':
+            response = self._response_from_object(message['r'])
             return ('page_crawled', response)
-        if message[b'type'] == b'request_error':
-            request = self._request_from_object(message[b'r'])
-            return ('request_error', request, to_native_str(message[b'error']))
-        if message[b'type'] == b'update_score':
-            return ('update_score', self._request_from_object(message[b'r']), message[b'score'], message[b'schedule'])
-        if message[b'type'] == b'add_seeds':
+        if message['type'] == 'request_error':
+            request = self._request_from_object(message['r'])
+            return ('request_error', request, message['error'])
+        if message['type'] == 'update_score':
+            return ('update_score', self._request_from_object(message['r']), message['score'], message['schedule'])
+        if message['type'] == 'add_seeds':
             seeds = []
-            for seed in message[b'seeds']:
+            for seed in message['seeds']:
                 request = self._request_from_object(seed)
                 seeds.append(request)
             return ('add_seeds', seeds)
-        if message[b'type'] == b'new_job_id':
-            return ('new_job_id', int(message[b'job_id']))
-        if message[b'type'] == b'offset':
-            return ('offset', int(message[b'partition_id']), int(message[b'offset']))
+        if message['type'] == 'new_job_id':
+            return ('new_job_id', int(message['job_id']))
+        if message['type'] == 'offset':
+            return ('offset', int(message['partition_id']), int(message['offset']))
         return TypeError('Unknown message type')
 
     def decode_request(self, message):
-        obj = dict_to_bytes(super(Decoder, self).decode(message))
-        return self._request_model(url=to_native_str(obj[b'url']),
-                                   method=obj[b'method'],
-                                   headers=obj[b'headers'],
-                                   cookies=obj[b'cookies'],
-                                   meta=obj[b'meta'])
-
+        obj = _convert_from_saved_type(super(Decoder, self).decode(message))
+        return self._request_model(url=obj['url'],
+                                   method=obj['method'],
+                                   headers=obj['headers'],
+                                   cookies=obj['cookies'],
+                                   meta=obj['meta'])
--- a/frontera/contrib/backends/remote/codecs/msgpack.py
+++ b/frontera/contrib/backends/remote/codecs/msgpack.py
@@ -2,15 +2,37 @@
 """ A MsgPack codec for Frontera. Implemented using native msgpack-python library.
 """
 from __future__ import absolute_import
-
+import logging
 from msgpack import packb, unpackb
 
 from frontera.core.codec import BaseDecoder, BaseEncoder
+import six
 from w3lib.util import to_native_str
 
 
+logger = logging.getLogger(__name__)
+
+
 def _prepare_request_message(request):
-    return [request.url, request.method, request.headers, request.cookies, request.meta]
+    def serialize(obj):
+        """Recursively walk object's hierarchy."""
+        if isinstance(obj, (bool, six.integer_types, float, six.binary_type, six.text_type)):
+            return obj
+        elif isinstance(obj, dict):
+            obj = obj.copy()
+            for key in obj:
+                obj[key] = serialize(obj[key])
+            return obj
+        elif isinstance(obj, list):
+            return [serialize(item) for item in obj]
+        elif isinstance(obj, tuple):
+            return tuple(serialize([item for item in obj]))
+        elif hasattr(obj, '__dict__'):
+            return serialize(obj.__dict__)
+        else:
+            logger.warning('unable to serialize object: {}'.format(obj))
+            return None
+    return [request.url, request.method, request.headers, request.cookies, serialize(request.meta)]
 
 
 def _prepare_response_message(response, send_body):
@@ -22,29 +44,29 @@ class Encoder(BaseEncoder):
         self.send_body = True if 'send_body' in kw and kw['send_body'] else False
 
     def encode_add_seeds(self, seeds):
-        return packb([b'as', [_prepare_request_message(seed) for seed in seeds]], use_bin_type=True, encoding="utf-8")
+        return packb([b'as', [_prepare_request_message(seed) for seed in seeds]], use_bin_type=True)
 
     def encode_page_crawled(self, response):
-        return packb([b'pc', _prepare_response_message(response, self.send_body)], use_bin_type=True, encoding="utf-8")
+        return packb([b'pc', _prepare_response_message(response, self.send_body)], use_bin_type=True)
 
     def encode_links_extracted(self, request, links):
         return packb([b'le', _prepare_request_message(request), [_prepare_request_message(link) for link in links]],
-                     use_bin_type=True, encoding="utf-8")
+                     use_bin_type=True)
 
     def encode_request_error(self, request, error):
-        return packb([b're', _prepare_request_message(request), str(error)], use_bin_type=True, encoding="utf-8")
+        return packb([b're', _prepare_request_message(request), str(error)], use_bin_type=True)
 
     def encode_request(self, request):
-        return packb(_prepare_request_message(request), use_bin_type=True, encoding="utf-8")
+        return packb(_prepare_request_message(request), use_bin_type=True)
 
     def encode_update_score(self, request, score, schedule):
-        return packb([b'us', _prepare_request_message(request), score, schedule], use_bin_type=True, encoding="utf-8")
+        return packb([b'us', _prepare_request_message(request), score, schedule], use_bin_type=True)
 
     def encode_new_job_id(self, job_id):
-        return packb([b'njid', int(job_id)], use_bin_type=True, encoding="utf-8")
+        return packb([b'njid', int(job_id)], use_bin_type=True)
 
     def encode_offset(self, partition_id, offset):
-        return packb([b'of', int(partition_id), int(offset)], use_bin_type=True, encoding="utf-8")
+        return packb([b'of', int(partition_id), int(offset)], use_bin_type=True)
 
 
 class Decoder(BaseDecoder):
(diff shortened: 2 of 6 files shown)


voith:
Build is still failing. need https://github.com/scrapinghub/frontera/pull/235 to fix the build error.

voith:
managed to get the build passing by temporarily merging #235. Here is the build https://travis-ci.org/scrapinghub/frontera/builds/183326040
The msg-pack and json codecs have been fixed. What is remaining now is to run an integration test manually on the Hbase Backend.

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -3,29 +3,59 @@
...
 import json
+import six
 from base64 import b64decode, b64encode
 from frontera.core.codec import BaseDecoder, BaseEncoder
-from w3lib.util import to_unicode, to_native_str
-from frontera.utils.misc import dict_to_unicode, dict_to_bytes
+from w3lib.util import to_unicode, to_bytes
+
+
+def _convert(obj):
```
sibiryakov: this is not entirely about conversion, but encoding of types preservation, I propose to change the name to `_encode_recursively()`

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -3,29 +3,59 @@
...
+    if isinstance(obj, bytes):
+        return 'bytes', to_unicode(obj)
+    elif isinstance(obj, dict):
+        return 'dict', [(_convert(k), _convert(v)) for k, v in six.iteritems(obj)]
+    elif isinstance(obj, (list, tuple)):
+        return type(obj).__name__, [_convert(item) for item in obj]
+    return 'other', obj
+
+
+def _reconvert(obj):
```
sibiryakov: the same here

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -6,31 +6,11 @@
...
-                obj[key] = serialize(obj[key])
-            return obj
-        elif isinstance(obj, list):
-            return [serialize(item) for item in obj]
-        elif isinstance(obj, tuple):
-            return tuple(serialize([item for item in obj]))
-        elif hasattr(obj, '__dict__'):
-            return serialize(obj.__dict__)
-        else:
-            return None
```
sibiryakov: This function skips silently the object if it's not possible to serialize. I propose to keep this behavior, but make it throw a warning.

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -42,28 +22,29 @@ def __init__(self, request_model, *a, **kw):
         self.send_body = True if 'send_body' in kw and kw['send_body'] else False
 
     def encode_add_seeds(self, seeds):
-        return packb([b'as', [_prepare_request_message(seed) for seed in seeds]])
+        return packb([b'as', [_prepare_request_message(seed) for seed in seeds]], use_bin_type=True, encoding="utf-8")
```
sibiryakov: according to short msgpack-python documentation, it's recommended to use encoding argument for ``unpackb`` not for ``packb``

sibiryakov: https://pypi.python.org/pypi/msgpack-python

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -45,6 +75,10 @@ def __init__(self, request_model, *a, **kw):
         self.send_body = kw.pop('send_body', False)
         super(Encoder, self).__init__(request_model, *a, **kw)
 
+    def encode(self, obj):
+        converted = _convert(obj)
```
sibiryakov: Currently, only meta requires type information to be stored along with the content. Type information cause message size growth, and introduces cpu overhead on parsing. So, I suggest to reimplement this to make it efficient, the same is (and especially) true for msgpack. I don't care much about json codec performance, but msgpack is very important one to be efficient, because it's used in large deployments.

tests/test_codecs.py
```diff
@@ -20,7 +24,8 @@ def check_request(req1, req2):
 
     enc = encoder(Request, send_body=True)
     dec = decoder(Request, Response)
-    req = Request(url="http://www.yandex.ru",method=b'GET', meta={b"test": b"shmest"}, headers={b'reqhdr': b'value'})
+    req = Request(url="http://www.yandex.ru", method=b'GET',
```
sibiryakov: I suggest to explicitly check the type for string objects in meta, to prove that your update codecs code really respects the type and preserves it.

sibiryakov:
also @Preetwinder please have a look. Your opinion would much appreciated.

frontera/contrib/backends/remote/codecs/json.py
voith: I agree. Will do.

frontera/contrib/backends/remote/codecs/json.py
voith: will do

frontera/contrib/backends/remote/codecs/json.py
voith: Won't this error occur for headers, cookies or in fact all dict type objects? Right now issues are only reported for meta.

Also any suggestions to make this efficient for msgpack?

frontera/contrib/backends/remote/codecs/json.py
voith: done

frontera/contrib/backends/remote/codecs/json.py
voith: done

frontera/contrib/backends/remote/codecs/json.py
voith: This is something I haven't fixed yet. @Preetwinder Your input here would be of great value

frontera/contrib/backends/remote/codecs/msgpack.py
voith: I have added this function back except with a minor change. I have excluded the encoding part for `six.text_type` as this will be handled now by msgpack

voith:
@Preetwinder Since this PR overwrites some of your changes, your input here would be highly appreciated.

chuckus:
@voith Thanks for your hard work! I've reviewed the code and it looks good, like the idea of using tuples for the encoding to reduce the payload size as opposed to a dict (or at least that's what I think you're doing).

voith:
@chuckus Thanks for your feedback. My idea to use tuples was to create a generator, But unfortunately, `JSONEncoder`  doesn't support generators, so I had to use lists in some places. So the code doesn't really achieve the desired optimization.

voith:
rebased. @sibiryakov Can I get a second review? I have still not implemented your suggestion for optimization. I'm waiting for your reply on the comments that I have left behind in the code review

frontera/contrib/backends/remote/codecs/msgpack.py
```diff
@@ -6,11 +6,29 @@
...
+                obj[key] = serialize(obj[key])
+            return obj
+        elif isinstance(obj, list):
+            return [serialize(item) for item in obj]
+        elif isinstance(obj, tuple):
+            return tuple(serialize([item for item in obj]))
+        elif hasattr(obj, '__dict__'):
+            return serialize(obj.__dict__)
+        else:
+            return None
```
sibiryakov: let's make it generate warning, that object of some non-serializable type was skipped.

tests/test_codecs.py
```diff
@@ -5,12 +5,38 @@
...
 from frontera.contrib.backends.remote.codecs.json import (Encoder as JsonEncoder,
                                                           Decoder as JsonDecoder,
-                                                          _convert, _reconvert)
+                                                          _encode_recursively, _decode_recursively)
 from frontera.contrib.backends.remote.codecs.msgpack import Encoder as MsgPackEncoder, Decoder as MsgPackDecoder
 from frontera.core.models import Request, Response
 import pytest
 
 
+def _compare_dicts(dict1, dict2):
```
sibiryakov: is there any reason to have such complex machinery for comparing dicts?

tests/test_codecs.py
```diff
@@ -5,12 +5,38 @@
...
+
+    shared_keys = set(dict2.keys()) & set(dict2.keys())
+
+    if not (len(shared_keys) == len(dict1.keys()) and len(shared_keys) == len(dict2.keys())):
+        return False
+
+    dicts_are_equal = True
+    for key in dict1.keys():
+        if type(dict1[key]) is dict:
+            dicts_are_equal = dicts_are_equal and _compare_dicts(dict1[key], dict2[key])
```
sibiryakov: here you could return False earlier in case of inequality, there is no need to iterate till the end

sibiryakov:
otherwise looks good!

tests/test_codecs.py
voith: Just wanted to be a little bit precise

Preetwinder:
LGTM.

voith:
@Preetwinder Thanks for your review. @sibiryakov Is there anything else that needs to be done to get this merged?

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -101,52 +135,51 @@ def __init__(self, request_model, response_model, *a, **kw):
         super(Decoder, self).__init__(*a, **kw)
 
     def _response_from_object(self, obj):
-        url = to_native_str(obj[b'url'])
+        url = obj['url']
         request = self._request_model(url=url,
```
Preetwinder: Are the keys guaranteed to be native strings for this and all the following `obj` accesses?

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -3,29 +3,59 @@
...
+    recursively encodes an object to unicode and store its type
+    """
+    if isinstance(obj, bytes):
+        return 'bytes', to_unicode(obj)
+    elif isinstance(obj, dict):
+        return 'dict', [(_encode_recursively(k), _encode_recursively(v)) for k, v in six.iteritems(obj)]
+    elif isinstance(obj, (list, tuple)):
+        return type(obj).__name__, [_encode_recursively(item) for item in obj]
+    return 'other', obj
+
```
Preetwinder: Isn't storing the types 'other, and 'dict' redundant, since the JSON encoder will take care of these types?

voith:
@Preetwinder thank you for your review. I will address your concerns tonight

voith:
sorry for keeping this PR on hold. Have been a little busy lately

sibiryakov:
NP, I had vacations too ;)

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -3,29 +3,59 @@
...
 from base64 import b64decode, b64encode
 from frontera.core.codec import BaseDecoder, BaseEncoder
-from w3lib.util import to_unicode, to_native_str
-from frontera.utils.misc import dict_to_unicode, dict_to_bytes
+from w3lib.util import to_unicode, to_bytes
+
+
+def _encode_recursively(obj):
+    """
+    recursively encodes an object to unicode and store its type
```
kmike: a nitpick: you can't encode to unicode because unicode is not an encoding (at least in Python), you can only decode to unicode

voith: @kmike Thanks for the review. 
I had previously named this method `convert`: https://github.com/scrapinghub/frontera/pull/233/commits/33347d9a9f0c5a9e732612350b320ca09309baa8#diff-6cd48fdd254557ffc1b50d60a0323364R12. 

Side Note: I think I should start contributing to scrapy. There's a lot I can learn from your reviews :)

frontera/contrib/backends/remote/codecs/json.py
voith: A dict can can have nested dicts or other  kinds of objects. Since we are doing some kind of conversion here, we want to save the type here for successful de-conversion.  Json Encoder taking care is not sufficient. We need to successfully Decode this to the original form for which we need to save type.
Although this implementation currently only converts strings, list and dicts it can be easily extended for something like sets. 
If you have any other implementation in mind, please let me know.

frontera/contrib/backends/remote/codecs/json.py
voith: have a look at this line https://github.com/scrapinghub/frontera/pull/233/files?diff=split#diff-6cd48fdd254557ffc1b50d60a0323364R154. The previous method called in the decoder was `dict_to_bytes` which was changing the native_string types to bytes. But the encoder was never changed and it used native_Strings. Since the encoder takes care that it uses native strings the decoder need not worry about any native_string conversion.

voith:
I've also tested this PR manually with the Hbase backend and it works fine for me

voith:
@sibiryakov Please don't merge this. I have to apply kmikes suggestion

tests/test_codecs.py
sibiryakov: you should understand that having such deep comparisons would require extra processor cycles, and in Frontera we tend to save them and avoid doing unnecessary computation if possible. so please think again if this is really needed. may be try to find an example.

frontera/contrib/backends/remote/codecs/json.py
sibiryakov: Still it's not clear why 'dict' and 'other' is needed? Please be more specific. My concern is connected with integration with other components. Imagine if somebody else decide to implement this codec in other language. The less type-related information we store here, the easier it would be.

frontera/contrib/backends/remote/codecs/json.py
sibiryakov: If url had type 'bytes' then after decoding it will have type bytes, and in Frontera URLs have to be native strings (when in process).

sibiryakov:
I think we're close. Let's clarify all the issues and try to merge it.

voith:
@sibiryakov you're concerns are real. I try to address these issues when I get some more time. Finding it hard to manage time! But i'll get back to this as soon as possible

frontera/contrib/backends/remote/codecs/json.py
voith: > If url had type 'bytes' then after decoding it will have type bytes

You are right about the fact that this PR will save the type of the object it was originally given to the codec and on on decoding it will return  the original type of value.

> and in Frontera URLs have to be native strings

If this is the case than such type checking and conversion should be done somewhere else, Not In a codec. 

>  so maybe we still need a preventive conversion to native strings

A codec should never change the type of the object it was originally inputed with. Forcing such a mechanism in codec is a hacky way of getting around this problem.

voith:
@sibiryakov I have tried my best to address your comments as Best as I possibly could. Please let me know if there's still any doubt. A a bit late here now(IST), so can't apply kmikes suggestion. But I hope I have atleast addressed your concerns.

frontera/contrib/backends/remote/codecs/json.py
sibiryakov: > A codec should never change the type of the object it was originally inputed with. Forcing such a mechanism in codec is a hacky way of getting around this problem.

Agree, it's not a codec's responsibility. But then we would need to check the rest of the code and make sure we're using native strings for URLs. URLs could appear in crawling strategy (sw), backend or message bus. Because such a change could break the code relying on URLs present as native strings.

sibiryakov: Ok, false alarm probably
https://github.com/scrapinghub/frontera/blob/master/frontera/core/models.py#L28
any thoughts @Preetwinder ?

frontera/contrib/backends/remote/codecs/json.py
sibiryakov: Yes, thanks for the explanation, now it's much clearer. Ok, let's leave the dict there. But could you leave the comment in the code explaining why it is, at least in short form.

sibiryakov:
Have you tried the crawling in distributed backends mode with Json codec @voith ?

voith:
@sibiryakov No I haven't. I only checked that its not blowing up anything. But I have tested msg-pack in distributed mode. I will let you know when I finish testing the json codec in distributed mode

frontera/contrib/backends/remote/codecs/json.py
Preetwinder: Yes, URLs shouldn't be a problem, they are guaranteed to be native strings. @voith I understand. The reason for using dict_to_bytes was that the JSON encoder returns unicode strings and we wanted meta, headers, etc to be bytes. But it seems the new function takes care of that.

frontera/contrib/backends/remote/codecs/json.py
Preetwinder: Thanks for the explanation, I get it.

voith: @sibiryakov will add a docstring.

voith:
I have changed the names of the methods `_encode_recursively` and  `_decode_recursively` as pointed out by @kmike. I have also added a docstring explaining in brief  the purpose of these methods and their design considerations.

But I didn't really get the time to test json-codec with the distributed backend. If anyone in the community could help me test it than it would be great. In any case it'll take some more time to test this thoroughly.

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -9,33 +9,49 @@
...
-        return 'dict', [(_encode_recursively(k), _encode_recursively(v)) for k, v in six.iteritems(obj)]
+        return 'dict', [(_convert_and_save_type(k), _convert_and_save_type(v)) for k, v in six.iteritems(obj)]
     elif isinstance(obj, (list, tuple)):
-        return type(obj).__name__, [_encode_recursively(item) for item in obj]
+        return type(obj).__name__, [_convert_and_save_type(item) for item in obj]
     return 'other', obj
 
 
-def _decode_recursively(obj):
+def _revert_from_saved_type(obj):
```
sibiryakov: IMO revert is used in the wrong context here http://dictionary.cambridge.org/dictionary/english/revert-to-sth, convert_from_saved_type would be better.

voith: will do

sibiryakov:
OK. Finall call before I merge it. @voith please rename the revert_* method name.
Anyone here has any ideas what it could break @lopuhin @kmike @Preetwinder ?

tests/test_codecs.py
lopuhin: For me it was not immediately obvious, but I think the main difference from built-in dict comparison is this line https://github.com/scrapinghub/frontera/pull/233/files#diff-252cc4202700183e8f0c782fb28dbfc2R34, so it will catch e.g. python 2 dicts with equal str and unicode values, int/float, etc.

frontera/contrib/backends/remote/codecs/json.py
voith: renamed

frontera/contrib/backends/remote/codecs/json.py
```diff
@@ -3,29 +3,75 @@
...
+    """
+    :param obj: object returned by `_convert_and_save_type`
+
+    Reverts the original state of the object converted
+    earlier by `_convert_and_save_type`. This method considers every
+    first element of the nested tuple as the original type information and
+    the second value to be the converted value. It applies the original type
+    recursively on the object to retrieve the original form of the object.
+    """
```
lopuhin: This is super minor, but I would maybe add a check that ``obj`` has exactly two items, e.g. with ``obj_type, obj_value =  obj``, just in case some list is passed instead of an object returned by ``_convert_and_save_type``

voith: @lopuhin thanks for the review. Please see if [this](https://github.com/scrapinghub/frontera/pull/233/commits/b7f5a2dc3a2d62e5f5940e486e756835aeb8d47e) looks good to you

lopuhin: Great, thanks @voith! 👍

voith:
@sibiryakov Here's a suspicious [line](https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/scrapy/converters.py#L84) I found out. I haven't really tested it, but I have a feeling that this will break. I think as part of this PR some clean up is also needed. All strings related to scrapy should be native strings.

Also I did try to test json codec with the distributed backend. But I was having some problems with my hbase setup. I have recently shifted to mac and `hbase-docker` doesn't seem to work well with it. Allow me some time to work on this. But I can't promise anything. Or if anyone else has the time than fell free to continue this PR.

sibiryakov:
Yeah @voith I noticed that. But that was broken before you, and I don't see how it is connected to this PR.

voith:
@sibiryakov You're probably right. I wasn't entirely sure. But give me a day to manually test everything. I managed to setup hbase locally. I will verify and let you know

sibiryakov:
Sure @voith

voith:
@sibiryakov I have finished testing json codec in distributed mode. I can confirm that everything works. The only error I was getting has been fixed in #252. If i apply that patch to this branch than everything works fine. Also that patch has nothing to do with this PR. That bug existed before I made this PR. This PR should be merged after #252 is merged.

voith:
I expect some merge conflicts in this branch if #252 gets merged first. I will resolve them later

sibiryakov:
ok, rebase it pls.

voith:
@sibiryakov I have rebased the PR and fixed merge conflicts.
I have also squashed the commits into two: one for msg-pack and the other for json

sibiryakov:
Congratulations and thank you @voith @Preetwinder and others! 🍺 🍺 🍕

voith:
Thanks to everyone who reviewed this PR and helped me make the code better! :)

sibiryakov:
thank you for bearing all this!

voith:
No problem @sibiryakov. I personally learnt a lot from this PR. 
 In fact if any further issues arise due to this PR, I will fix them myself.

Issue #102, opened 2016-01-27
Get rid of OverusedBuffer 


sibiryakov:
Buffering requests to busy hosts should be responsibility of fetcher component. We need to figure out how to change interfaces, and how to support necessary buffering logic in our default fetcher (Scrapy).

sibiryakov:
That would require talking directly with Scrapy Downloader, which isn't possible using Scrapy Scheduler interface.

PR #238, opened 2017-01-08
Move towards "Get rid of OverusedBuffer" (#102)

ZipFile:
* implement new interface for OverusedBuffer
* update examples
* provide fallback buffering logic to not to ruin existing behaviour

In general, this changeset targeted to ease pain when we'll decide to finally remove OverusedBuffer from the `frontera.core`.


sibiryakov:
Hi,
my feelings are kind-of divided. There's a lot of work, and the code quality is relatively high. However, I see different the removal of OverusedBuffer.

The reason why OverusedBuffer is needed, is imperfection of crawling strategies. It's happening very often that crawling strategy is written in a way that it's causing flooding of spider with thousands of queries to the same website. Thus crawler either waits for delays (polite crawl) or waits for overloaded website responses wasting time. The time between these delays could be used to query other websites.

So overused buffer accumulates requests to a single website to prevent getting them into the main downloader queue when it's busy. If they get there, then it will force downloader to process only requests from one or few domains, therefore causing waste of time and resources (usually CPU) under usage. For example Scrapy downloader is limited by CONCURRENT_REQUESTS* settings. Any other downloader could behave slightly different, but when flooded with a single or few domains, it has nothing to do, but process them. There are two ways to solve that problem:
    a) provide the list of busy domains on every attempt to request a new batch,
    b) avoid generating a batch with number of distinct domains less than N.

When dealing with large broad crawls we recommend to use HBaseBackend so b) will be the main strategy, and a) will be used to add additional level of protection. In case Frontera is used with SQLAlchemyBackend only a) is used.

The only situation when OverusedBuffer could be removed is moving it to Scrapy downloader, but that way we will 1) introduce additional requirements for downloaders to behave efficient and 2) would need to change the way downloading/queueing within Scrapy works. I believe there is a value in it for non-Scrapy downloaders.

To summarize, I'm really not sure if we there is a need to make OverusedBuffer a Scrapy-only component.

ZipFile:
> The reason why OverusedBuffer is needed, is imperfection of crawling strategies. It's happening very often that crawling strategy is written in a way that it's causing flooding of spider with thousands of queries to the same website. Thus crawler either waits for delays (polite crawl) or waits for overloaded website responses wasting time. The time between these delays could be used to query other websites.

If I understand Frontera's architecture right, strategy is all about _what_ to crawl, not _how_. So that's not a strategy fault to begin with. Intense flood problem need to be solved somewhere else.

> There are two ways to solve that problem:
> a) provide the list of busy domains on every attempt to request a new batch,
> b) avoid generating a batch with number of distinct domains less than N.

Talking back to the strategy (a) neither a good idea too, it will overcomplicate entire thing. I believe, altering dequeue logic (b) is a way to go.

>The only situation when OverusedBuffer could be removed is moving it to Scrapy downloader, but that way we will 1) introduce additional requirements for downloaders to behave efficient and 2) would need to change the way downloading/queueing within Scrapy works. I believe there is a value in it for non-Scrapy downloaders.

I'm not familiar with scrapy, but I recall they got some improvements on [customizing queuing](https://doc.scrapy.org/en/latest/topics/settings.html#std:setting-SCHEDULER_PRIORITY_QUEUE) lately. Maybe it makes sense to adjust OverusedBuffer api to match queuelib's interface and move it to `frontera.contrib`?

> To summarize, I'm really not sure if we there is a need to make OverusedBuffer a Scrapy-only component.

In fact, I highly disagree about current OverusedBuffer location, that's what made me to produce this PR. It feels wrong to do such kind of buffering at message bus backend. Right now, if you call the `get_next_requests` there is no guarantee that attempt to fetch the data will be performed, because it might remain in the hidden buffer. Moreover, different spider implementation might have it's own buffering, so doing it twice is redundant.

Given the context of discussion, maybe manager would be ok to put the OverusedBuffer into, i.e. implement either `OverusedBufferMixin`  or rather `OverusedBufferManager`?

Anyways, this PR is kind of proposal. I can split it into few smaller PRs if you think some of my code would be useful right now.

sibiryakov:
Hi @ZipFile It's great that you have a strong interest in this problem 👍  I'll try to address your arguments below.

> If I understand Frontera's architecture right, strategy is all about what to crawl, not how. So that's not a strategy fault to begin with. Intense flood problem need to be solved somewhere else.
That was our initial idea. But after few months of development we realized these problems are tightly coupled. See yourself: there's a batch concept, so spiders are always consuming batches of URLs from the queue. Therefore, we need to prepare batch with multiple domains, instead of a sequencing the whole domain one by one. This making us designing the queue the way it could return pages from multiple domains, irrelevant to the way they were added to the queue. This queue have to scale to terabytes of requests. And this is already Frontera's territory: backend abstraction. Therefore it all end-up as a framework with a set of primitives for building scalable web crawling applications. Another reason is lack of queueing in open source downloaders: some have frontier coupled with fetcher, others are not designed for massive amount of requests (requests), and scrapy has it's own queueing model. We had to find a way to play together with all that variability of approaches.

sibiryakov:
> Talking back to the strategy (a) neither a good idea too, it will overcomplicate entire thing. I believe, altering dequeue logic (b) is a way to go.

I wish it to be like that. Consider use case when there are 10 websites and they all of different size (orders of magnitude). There is no other option than going a) at the beginning because the queue contents will not be balanced because of natural size of websites.

sibiryakov:
> I'm not familiar with scrapy, but I recall they got some improvements on customizing queuing lately. Maybe it makes sense to adjust OverusedBuffer api to match queuelib's interface and move it to frontera.contrib?

I think it's not connected. Currently Frontera is integrating by means of custom scheduler and set of of Middlewares. The request for new batch comes from scheduler and is propagated to Frontera's manager, and there is quite fragile reading of downloader state before the actual batch preparation starts in Frontera. Also there is another useless buffering happens in Frontera's scheduler, because Scrapy's scheduler API is designed to consume a single request with one call. I believe Scrapy and Frontera API changes should address these issues.

sibiryakov:
> In fact, I highly disagree about current OverusedBuffer location, that's what made me to produce this PR. It feels wrong to do such kind of buffering at message bus backend. Right now, if you call the get_next_requests there is no guarantee that attempt to fetch the data will be performed, because it might remain in the hidden buffer. Moreover, different spider implementation might have it's own buffering, so doing it twice is redundant.

Our goal here is to ease the setup for majority of users. Most of the time when people are dealing with MessageBusBackend is dealing with large volumes so they decided to distribute the crawl. Therefore OverusedBuffer is added there by default. Of course it would be nice to make it transparent on how Backend works, but that's not the primary goal.

Buffering and batch operation in spider process is needed to avoid delays when processing requests. Calling any messaging system API isn't as fast as reading data from memory.

sibiryakov:
thanks for bringing this up.

ZipFile:
Ok, I see. So the problem is that large website will have higher priority due amount of the unique urls. But our goal is to make it more democratic? In this case, entire overused buffer concept is wrong. Extremely large OverusedBuffer needed to deal with it in somehow efficient manner now. Moreover, there is also need to deal with data loss (i.e. spider crash) so overused buffer here might introduce weird gaps between scheduled and crawled urls. Sure, production-grade code won't crash out of blue, but there is still plenty of ways to loss the buffer.

sibiryakov:
I don't understand about democracy ;) But, I do agree with all your other points. Moreover, we're already facing OOM kills because of (presumably) large OverusedBuffer contents and plan to submit fixes.

sibiryakov:
I would appreciate if you could have an ideas on how to address the above mentioned issues.

PR #241, opened 2017-01-16
[RFC] Better OverusedBuffer implementation

ZipFile:
Well, why not? Won't harm while we're in search of the proper solution.

Tests located at `tests/test_core_overused_buffer.py`.

### DIFF
--- a/frontera/core/__init__.py
+++ b/frontera/core/__init__.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from six.moves.urllib.parse import urlparse
 from socket import getaddrinfo
-from collections import deque
+from collections import defaultdict, deque
 import six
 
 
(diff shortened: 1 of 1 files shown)


sibiryakov:
We need a test case for this.

sibiryakov:
Well done!
If you're looking around OverusedBuffer let's think about reliability. There is no protection from excessive memory consumption. I propose to introduce some setting designating memory limit of OverusedBuffer and another option selecting the type of behavior when memory consumption has reached the limit.
I see two possible strategies:
 - Dropping requests. We should do it the balanced way, e.g. not to drop a single domain, rather drop equal amount from all domains, always logging with WARN level how much was dropped and with DEBUG what exactly was dropped.
 - Stop querying storage backend. Again logging that memory limit was reached, and OB is now using only it's internal storage.

let me know what do you think

ZipFile:
Hello.

Sounds wise enough. I'd like to add third behaviour: do nothing (for debugging purposes). But I think it's task for separate PR.

Also there is a problem how to find memory usage. Should it be like "max n requsts per host in a buffer" or should it be something that mess with python internals like 'sys.getsizeof'?

sibiryakov:
> Also there is a problem how to find memory usage. Should it be like "max n requsts per host in a buffer" or should it be something that mess with python internals like 'sys.getsizeof'?
I think first one, or it could be max number of requests in OB.

sibiryakov:
what [rfc] in the title means?

voith:
Request for Comments?

sibiryakov:
Hey @ZipFile is it ready for merge?

ZipFile:
Yeah, sure.

sibiryakov:
Thanks 🍺 
do you think it's going to be more efficient from performance prospective @ZipFile ?

ZipFile:
There won't be lots of gain, but overall performance should improve a bit.

According to [research on the net ](http://stackoverflow.com/a/12556053) inserting items to the buffer will definitely improve. As of popping up from the buffer, I believe, removing extra calls and checks might do the trick.

PR #244, opened 2017-01-19
Don’t change scrapy.settings.default_settings.

### LINKED ISSUE #242: simplify examples of Scrapy settings
In frontera docs and examples Scrapy middlewares are added like this:

```py
from scrapy.settings.default_settings import SPIDER_MIDDLEWARES, DOWNLOADER_MIDDLEWARES

SPIDER_MIDDLEWARES.update({
    'frontera.contrib.scrapy.middlewares.schedulers.SchedulerSpiderMiddleware': 999,
    'frontera.contrib.scrapy.middlewares.seeds.file.FileSeedLoader': 1,
})
DOWNLOADER_MIDDLEWARES.update({
    'frontera.contrib.scrapy.middlewares.schedulers.SchedulerDownloaderMiddleware': 999,
})
```

Is this a requirement? Usually users should just define their settings like this, without any imports:

```py
SPIDER_MIDDLEWARES = {
    'frontera.contrib.scrapy.middlewares.schedulers.SchedulerSpiderMiddleware': 999,
    'frontera.contrib.scrapy.middlewares.seeds.file.FileSeedLoader': 1,
}
DOWNLOADER_MIDDLEWARES = {
    'frontera.contrib.scrapy.middlewares.schedulers.SchedulerDownloaderMiddleware': 999,
}
```

Changing objects from `scrapy.settings.default_settings` directly is not supported by Scrapy, there are no docs/tests for this, and it may break in subtle ways. Scrapy settings are not global, but here a global variable is changed (the one which is used as an initial value of settings).

For example, if there are  two settings files, you set SCRAPY_SETTINGS_MODULE=settings1.py env variable to use `settings1.py` as settings, then run `scrapy crawl foo`. It tries to find a spider `foo` - to do so it imports python files, `settings2.py` get imported, global `scrapy.settings.default_settings.DOWNLOADER_MIDDLEWARES` is patched ==> you get options from `settings2.py` applied, not only from `settings1.py`.

kmike:
I haven't changed some of the examples in the docs because technically they are still correct: if you already have e.g. DOWNLOADER__MIDDLEWARES enabled you can use `.update({...})` to add frontera middlewares. All `scrapy.settings.default_settings` mentions are removed. It should fix #242.

### DIFF
--- a/docs/source/topics/cluster-setup.rst
+++ b/docs/source/topics/cluster-setup.rst
@@ -106,18 +106,17 @@ The logging can be configured according to https://docs.python.org/2/library/log
 6. Configure Scrapy settings module. It's located in Scrapy project folder and referenced in scrapy.cfg. Let's add
 there::
 
-    from scrapy.settings.default_settings import SPIDER_MIDDLEWARES, DOWNLOADER_MIDDLEWARES
-
     FRONTERA_SETTINGS = ''  # module path to your Frontera spider config module
 
     SCHEDULER = 'frontera.contrib.scrapy.schedulers.frontier.FronteraScheduler'
-    SPIDER_MIDDLEWARES.update({
+
+    SPIDER_MIDDLEWARES = {
         'frontera.contrib.scrapy.middlewares.schedulers.SchedulerSpiderMiddleware': 999,
         'frontera.contrib.scrapy.middlewares.seeds.file.FileSeedLoader': 1,
-    })
-    DOWNLOADER_MIDDLEWARES.update({
+    }
+    DOWNLOADER_MIDDLEWARES = {
         'frontera.contrib.scrapy.middlewares.schedulers.SchedulerDownloaderMiddleware': 999,
-    })
+    }
 
 
 Starting the cluster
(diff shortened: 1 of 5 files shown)


sibiryakov:
@kmike IMO, the problem you mentioned in #242 will not be completely solved if will not fix it in examples. People are tend to copy and modify them. If you have time, please fix them too.

kmike:
@sibiryakov copy-pasting is problematic with both styles. 

If there is no DOWNLOADER_MIDDLEWARES option defined in settings.py (this option is  commented out by default) then copy-pasting `DOWNLOADER_MIDDLEWARES = {}` would work, but copy-pasting `DOWNLOADER_MIDDLEWARES.update({...})` won't. 

If `DOWNLOADER_MIDDLEWARES` option is defined then copy-pasting `DOWNLOADER_MIDDLEWARES = {...}` could lead to a bug (a value will be overwritten), but copy-pasting `DOWNLOADER_MIDDLEWARES.update({...})` would work fine. 

Either way, currently examples don't have `default_settings` import, and they tell users to update their DOWNLOADER_MIDDLEWARES, so they are technically correct.

sibiryakov:
Here are the imports
https://github.com/scrapinghub/frontera/blob/master/examples/cluster/bc/settings.py#L2
https://github.com/scrapinghub/frontera/blob/master/examples/general-spider/general/settings.py#L10

https://github.com/scrapinghub/frontera/blob/master/examples/scrapy_recording/scrapy_recording/settings.py - no import but update is used.

kmike:
@sibiryakov right, all of them are fixed in this PR.

sibiryakov:
oops) thank you :taco: !

PR #246, opened 2017-01-20
don’t run Travis tests multiple times for pull requests from branches

kmike:
For example, in https://github.com/scrapinghub/frontera/pull/244 Travis tests were executed twice. The fix is copy-pasted from scrapy config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: python
 python: 2.7
+branches:
+  only:
+    - master
+    - /^\d\.\d+$/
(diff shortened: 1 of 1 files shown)


sibiryakov:
@kmike I'm trying to find when they were executed twice, and I can't 
https://travis-ci.org/scrapinghub/frontera/builds
could you point me out?

kmike:
@sibiryakov see https://github.com/scrapinghub/frontera/pull/244:

sibiryakov:
I see, thanks @kmike

PR #247, opened 2017-01-26
Using iterable instead of sequence when fetching states

sibiryakov:
this will allow using more container types in users code.

### DIFF
--- a/frontera/worker/strategy.py
+++ b/frontera/worker/strategy.py
@@ -15,7 +15,7 @@ from twisted.internet.task import LoopingCall
 from twisted.internet import reactor
 
 from frontera.settings import Settings
-from collections import Sequence
+from collections import Iterable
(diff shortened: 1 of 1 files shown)


ZipFile:
Looks good.

frontera/worker/strategy.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, states):
         self._fingerprints = set()
 
     def to_fetch(self, requests):
-        if isinstance(requests, Sequence):
+        if isinstance(requests, Iterable):
             self._fingerprints.update([x.meta[b'fingerprint'] for x in requests])
```
ZipFile: `set.update()` receives iterable too. You could save bit of memory here converting list to generator (i.e. remove `[` and `]`).

PR #250, opened 2017-02-06
DBW hardening: don't skip the whole batch, instead a single message

sibiryakov:
Otherwise it stalls and keep processing the same batch from Kafka.

### DIFF
--- a/frontera/worker/db.py
+++ b/frontera/worker/db.py
@@ -137,51 +137,57 @@ class DBWorker(object):
                 logger.error("Decoding error: %s", e)
                 continue
             else:
-                type = msg[0]
-                if type == 'add_seeds':
-                    _, seeds = msg
-                    logger.info('Adding %i seeds', len(seeds))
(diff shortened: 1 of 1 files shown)


frontera/worker/db.py
```diff
@@ -137,51 +137,57 @@ def consume_incoming(self, *args, **kwargs):
...
+                            if lag < self.max_next_requests or offset == 0:
+                                self.spider_feed.mark_ready(partition_id)
+                            else:
+                                self.spider_feed.mark_busy(partition_id)
+                        continue
+                    logger.debug('Unknown message type %s', type)
+                except Exception as exc:
+                    logger.exception(exc)
+                    if logger.isEnabledFor(logging.DEBUG):
+                        logger.debug("Message caused the error %s", str(msg))
```
voith: Can we rephrase this as `An error occurred in the following message: %s`?

`Message caused the error` is a little confusing to me

PR #254, opened 2017-02-12
Fixed scheduler process_spider_output() to yield requests

### LINKED ISSUE #253: Scrapy's spider middlewares having `process_spider_output()` don't work with frontera
I noticed that scrapy's `RefererMiddleware` middleware was not setting the `referer` header for requests as expected while using frontera:

Here's an example of a spider to replicate the issue:
```python
from __future__ import print_function

import scrapy
from scrapy.http import Request


class QuotesSpider(scrapy.Spider):
    name = 'quotes'

    def parse(self, response):
        yield Request('http://quotes.toscrape.com/page/2/', callback=self.check_headers)

    def check_headers(self, response):
        print('headers=%s' % response.request.headers)

```
My seeds file contains this url: `http://quotes.toscrape.com/page/1/`. I tested this with frontera's `sqlalchemy.FIFO`

Here's a screenshot of my logs with frontera:
screen shot 2017-02-12 at 2 08 38 pm

Here's a screenshot with only scrapy(assuming I'm providing start_urls instead of seeds):
screen shot 2017-02-12 at 2 15 08 pm

<h1>Cause:</h1>

Here's what is causing the issue, read this [code](https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/scrapy/schedulers/frontier.py#L111-L114) to understand the cause of the bug.
The `process_spider_output` method in frontera's scheduler stores all requests but doesn't yield them. Frontera's  [SchedulerSpiderMiddleware](https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/scrapy/middlewares/schedulers.py#L17-L19) is what makes a call to the schedulers `process_spider_output()`. But since requests are not yielded by this middleware all middlewares following `SchedulerSpiderMiddleware` in the chain won't work as expected. eg: scrapy's `RefererMiddleware`.  Read the following [code](https://github.com/scrapy/scrapy/blob/master/scrapy/spidermiddlewares/referer.py#L22) for a better understanding.

The scrapy [docs](https://doc.scrapy.org/en/latest/topics/spider-middleware.html?highlight=process_spider_output#scrapy.spidermiddlewares.SpiderMiddleware.process_spider_output) say the following:
> `process_spider_output()` must return an iterable of Request, dict or Item objects. 

I don't know why frontera needed this, but this seems wrong to me. Frontera should not affect the functioning of scrapy.

voith:
fixes #253 
Here's a screenshot using the same code discussed [here](https://github.com/scrapinghub/frontera/issues/253#issue-207038554).
screen shot 2017-02-12 at 3 13 48 pm

Nothing seems to break when testing this change manually. The only [test](https://github.com/scrapinghub/frontera/blob/master/tests/test_frontera_scheduler.py#L121-L122) that was failing was wrong IMO because it passed a list of requests and items and was only expecting items in return. I have modified that test to make it compatible with this patch.

I've the split this PR into three commits:
- The first commit adds a test to reproduce the bug.
- The second commit fixes the bug
- The third commit fixes the broken test discussed above

<h3>A note about the tests added:</h3> 
The tests might be a little difficult to understand on the first sight. I would recommend to read the following code in order understand the tests:

- https://github.com/scrapy/scrapy/blob/master/scrapy/core/spidermw.py#L34-L73: This is to understand how scrapy processes the different methods of the spider middleware.
- https://github.com/scrapy/scrapy/blob/master/scrapy/core/scraper.py#L135-L147: This is to understand how the scrapy core executes the spider middleware methods and passes the control to the spider callbacks.

I have simulated the above discussed code in order to write the test.

### DIFF
--- a/frontera/contrib/scrapy/schedulers/frontier.py
+++ b/frontera/contrib/scrapy/schedulers/frontier.py
@@ -110,8 +110,7 @@ class FronteraScheduler(Scheduler):
         for element in result:
             if isinstance(element, Request):
                 links.append(element)
-            else:
-                yield element
+            yield element
         frontier_request = response.meta[b'frontier_request']
         self.frontier.page_crawled(response)  # removed frontier part from .meta
         # putting it back, to persist .meta from original request
--- a/tests/test_frontera_scheduler.py
+++ b/tests/test_frontera_scheduler.py
@@ -113,12 +113,18 @@ class TestFronteraScheduler(object):
     def test_process_spider_output(self):
         i1 = {'name': 'item', 'item': 'i1'}
         i2 = {'name': 'item', 'item': 'i2'}
+        no_requests = 3
         result = [r1, r2, r3, i1, i2]
         resp = Response(fr1.url, request=Request(fr1.url, meta={b'frontier_request': fr1}))
         crawler = FakeCrawler()
         fs = FronteraScheduler(crawler, manager=FakeFrontierManager)
         fs.open(Spider)
-        assert sorted(list(fs.process_spider_output(resp, result, Spider)), key=lambda i: sorted(i['item'])) == \
+        out = list(fs.process_spider_output(resp, result, Spider))
+        assert len(out) == len(result)
+        out_request = out[:no_requests]
+        assert set(r.url for r in out_request) == set(r.url for r in result[:no_requests])
+        out_items = out[no_requests:]
+        assert sorted(out_items, key=lambda i: sorted(i['item'])) == \
             sorted([i1, i2], key=lambda i: sorted(i['item']))
         assert isinstance(fs.frontier.manager.responses[0], FResponse)
         assert fs.frontier.manager.responses[0].url == resp.url
--- a/tests/test_scrapy.py
+++ b/tests/test_scrapy.py
@@ -1,11 +1,22 @@
 # -*- coding: utf-8 -*-
-
 from __future__ import absolute_import
-from frontera.contrib.scrapy.converters import RequestConverter, ResponseConverter
+
+import sys
+
+from scrapy.core.spidermw import SpiderMiddlewareManager
+from scrapy.http import Request, Response
 from scrapy.http.request import Request as ScrapyRequest
 from scrapy.http.response import Response as ScrapyResponse
-from frontera.core.models import Request as FrontierRequest
+from scrapy.spiders import Spider
+from scrapy.utils.test import get_crawler
+from twisted.internet.defer import Deferred
+from twisted.trial import unittest
 from w3lib.util import to_bytes
+
+from frontera.contrib.scrapy.converters import (RequestConverter,
+                                                ResponseConverter)
+from frontera.core.models import Request as FrontierRequest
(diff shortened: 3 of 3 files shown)


sibiryakov:
Hey @voith sorry for a long absence. It looks absolutely fine. I hope your complex tests work. Ready for merge?

voith:
Hi @sibiryakov, This is ready for merge. You can view that the test works by looking at the builds.
like always I add the regression test in the first commit to show failure and than add the fix in the second commit.

sibiryakov:
thank you! 🍺

isra17:
This PR broke Frontera behaviour. Now every yielded request end up in as a call to `add_seeds` as well. The initial behavior did swallow the requests in the middleware so Scrapy wasn't scheduling them. By yielding the request from the middleware, it now leave them to scrapy who eventually enqueue them with [`FronteraScheduler::enqueue_request `](https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/scrapy/schedulers/frontier.py#L91). Either fix `enqueue_request` or revert this. For the original issue, shouldn't have this been simply fixed by setting a lower priority to the `SchedulerSpiderMiddleware`?

isra17:
Ping @sibiryakov

voith:
@isra17 I too had noticed that `add_seeds` was getting called for every request yielded. I wasn't aware that it was my PR that broke this. 
You are right about the fact the problem can be solved by setting a lower priority, But I thought that the code added earlier was a mistake. I was expecting @sibiryakov to tell me what could possibly break with this change.
I tested the change I made in this PR and made sure that nothing broke, but did not anticipate that it would affect performance. 

Well frontera should have had a test case for this.

voith:
I've opened a PR to revert this change https://github.com/scrapinghub/frontera/pull/273

isra17:
Don't worry about that, this is not the kind of issue that broke vanilla Frontera in obvious manner. I didn't see it until I had a middleware with some logic specific to links_extracted.

sibiryakov:
this PR fixes the problem probably https://github.com/scrapinghub/frontera/pull/261,
the root of the problem is Scrapy Engine behavior which is calling enqueue_request() every time it gets request instance from one of downloader middlewares (but not always).

sibiryakov:
There is no need to revert this change, it's a step in the right direction: why should allow other middlewares to operate on objects passed Frontera middlewares.

isra17:
Unless I'm missing something, won't #261 end up scheduling twice the requests? 
`process_spider_output` is still calling `links_extracted` for all requests, then passing them to Scrapy which in his turn calls `enqueue_request` that itself add the request to the `_pending_requests`. So now the requests live both in the frontier queue and the spider queue.

sibiryakov:
@isra17 This will happen only if you have some middleware in Scrapy yielding all requests it gets. Normally, this shouldn't happen.

sibiryakov:
@isra17 I spend more time looking into this, and I think you're right: we will get requests in two places: temp. queue and frontier. I'll release the fix soon.

sibiryakov:
See https://github.com/scrapinghub/frontera/pull/276

PR #259, opened 2017-02-19
MessageBusBackend improvements

ZipFile:
* Split buffering and aggregation features onto separate mixins;
* minor style/readability improvements;
* tests rewritten.


frontera/contrib/backends/remote/messagebus.py
```diff
@@ -41,41 +45,41 @@ def frontier_stop(self):
...
-                request = self._decoder.decode_request(encoded)
-            except Exception as exc:
-                self._logger.warning("Could not decode message: {0}, error {1}".format(encoded, str(exc)))
-            else:
-                requests.append(request)
-        self.spider_log_producer.send(b'0123456789abcdef0123456789abcdef012345678',
-                                      self._encoder.encode_offset(self.partition_id,
-                                                                  self.consumer.get_offset(self.partition_id)))
-        return requests
+                yield self._decoder.decode_request(encoded)
```
sibiryakov: All users of this Backend have to be compatible with generators. Are they? Mainly Frontera Scrapy scheduler.

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -93,19 +97,65 @@ def states(self):
...
+        self._buffer = OverusedBuffer(
+            super_self.get_next_requests,
+            self._logger.debug,
+        )
+
+    def get_next_requests(self, max_n_requests, **kwargs):
+        return self._buffer.get_next_requests(max_n_requests, **kwargs)
+
+
+class AggregatePerHostMixin(object):
```
sibiryakov: Is GenericMessageBusBackend usable without AggregatePerHostMixin? Please describe the use case.

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -93,19 +97,65 @@ def states(self):
...
-        if b'fingerprint' not in link.meta[b'domain']:
-            continue
-        host_fprint = link.meta[b'domain'][b'fingerprint']
-        if host_fprint not in per_host:
-            per_host[host_fprint] = []
-        per_host[host_fprint].append(link)
+    per_host = defaultdict(list)
+
+    for request in requests:
+        key = get_host_fprint(request)
```
sibiryakov: The above code skips the link, if fingerprint wasn't found in meta[b'domain'], but your implementation adds it with None key, which will cause MessageBus to bark.

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -41,41 +45,41 @@ def frontier_stop(self):
         self.spider_log_producer.flush()
 
     def add_seeds(self, seeds):
-        per_host = aggregate_per_host(seeds)
-        for host_fprint, host_links in six.iteritems(per_host):
-            self.spider_log_producer.send(host_fprint,
-                                          self._encoder.encode_add_seeds(host_links))
+        key = b'0123456789abcdef0123456789abcdef012345678'
```
ZipFile: There is no response nor request to calculate fprint from. The idea was not to perform aggregation at this point.

In overall, I dislike `b'0123456789abcdef0123456789abcdef012345678'` thing too, same goes for sending offsets after get_next_requests, since user might want to use different fprint calculation function. Most likely, I miss something important about how our message bus works, would be great if you could provide your opinion on what should be done here.

frontera/contrib/backends/remote/messagebus.py
ZipFile: Got the point, will check this. I believe, I've seen somewhere that get_next_requests should be iterable, not just list. From my point of view, making it iterable might open path to different kind of asynchronous backends in the future.

frontera/contrib/backends/remote/messagebus.py
ZipFile: My use case is to use more complex aggregation than just by host. I need to aggregate both by client and by host in order to achieve my goals.

So yes, GenericMessageBusBackend won't be much use in distributed mode. It was supposed to provide basic blocks to build the whatever-you-want MB backends.

ZipFile:
Thanks for review, I'll try to update the code somewhere around weekend.

ZipFile:
New patchset arrived:

* Reverted change of the `get_next_requests` to generator. It appears, problem is much more complex than I expected. Fixing affected code is outside of this PR's scope;
* removed `GenericMessageBusBackend` with mixins in favor to configuration through manager settings;
* improved commit structure.

tox.ini
```diff
@@ -28,6 +28,7 @@ commands = flake8 setup.py frontera
 [flake8]
 ignore = E265,E501,F401,W391,W292,E226
 exclude = frontera/_version.py,versioneer.py,docs/source/conf.py,frontera/contrib/backends/opic/discovery.py
+application-import-names = frontera, tests
```
sibiryakov: why is this change needed?

frontera/contrib/backends/remote/messagebus.py
```diff
@@ -94,18 +117,20 @@ def states(self):
...
-        per_host[host_fprint].append(link)
+
     return per_host
 
 
 def get_host_fprint(request):
-    if b'fingerprint' not in request.meta[b'domain']:
+    try:
+        return request.meta[b'domain'][b'fingerprint']
+    except KeyError:
```
sibiryakov: Side note:
I think we should hard-code UrlFingerprintMiddleware and DomainMiddleware during initialization in Manager. Without them Frontera isn't operational. So it doesn't make any sense to check for presence of this data, or check if these middlewares are enabled.

frontera/settings/default_settings.py
```diff
@@ -26,6 +26,8 @@
 MAX_REQUESTS = 0
 MESSAGE_BUS = 'frontera.contrib.messagebus.zeromq.MessageBus'
 MESSAGE_BUS_CODEC = 'frontera.contrib.backends.remote.codecs.msgpack'
+MESSAGE_BUS_REQUEST_AGGREGATION_FUNCTION = 'frontera.contrib.backends.remote.messagebus.aggregate_per_host'
```
sibiryakov: these methods needs to be documented in frontera-settings.rst

tests/contrib/backends/remote/test_message_bus_backend.py
```diff
@@ -0,0 +1,328 @@
...
+    aggregate_per_host,
+    get_host_fprint,
+)
+from frontera.core.models import Request, Response
+from frontera.settings import Settings
+
+import tests
+
+
+patch = tests.mock.patch
```
sibiryakov: looks like that on is unused?

sibiryakov:
In general, it's OK. Only docs are needed.

tox.ini
ZipFile: Hints for `flake8-import-order`. Needed to suppress invalid import order errors.

frontera/contrib/backends/remote/messagebus.py
ZipFile: Agreed, that was not clear for me too at first. Frontera in distributed mode totally unusable without those two.

tests/contrib/backends/remote/test_message_bus_backend.py
ZipFile: that's right

frontera/settings/default_settings.py
ZipFile: Ok, will update the doc.

sibiryakov:
Ok waiting for the docs @ZipFile

ZipFile:
Bit late, but docs are updated now.

docs/source/topics/frontera-settings.rst
```diff
@@ -238,6 +238,25 @@ Default: ``frontera.contrib.backends.remote.codecs.msgpack``
...
+Defaults to per-host aggregating function.
+
+.. setting:: MESSAGE_BUS_USE_OVERUSED_BUFFER
+
+MESSAGE_BUS_USE_OVERUSED_BUFFER
+-------------------------------
+
+Default: ``True``
+
+Enables use of :term:`overused buffer` in :class:`MessageBusBackend <frontera.contrib.backends.remote.messagebus.MessageBusBackend>`.
```
sibiryakov: Is this term defined somewhere? If not please define it in glossary.rst

sibiryakov:
aggregation is outdated, this functionality must live in crawling strategy

sibiryakov:
optional overused buffer makes sense, but unlikely to be a problem in most cases. May be worth to live it as is, assuming that people who want to disable it are able to fork and customise Frontera.

PR #260, opened 2017-02-22
WIP Add priority when converting between scrapy<->frontier requests.

rmax:


sibiryakov:
Here we send too many fields which are None. Could we send fields only if they're not None? This will improve the performance of serialization.

rolando:
I updated the changes to set `scrapy_callback` and `scrapy_errback` only if set. Let me know if it looks OK and I will add tests.

frontera/contrib/scrapy/converters.py
```diff
@@ -19,24 +19,25 @@ def to_frontier(self, scrapy_request):
...
         if b'frontier_request' in scrapy_meta:
             request = scrapy_meta[b'frontier_request']
             if isinstance(request, FrontierRequest):
                 meta = request.meta
             del scrapy_meta[b'frontier_request']
 
         meta.update({
-            b'scrapy_callback': cb,
-            b'scrapy_errback': eb,
+            b'scrapy_priority': scrapy_request.priority,
```
sibiryakov: Let's do the same with priority? It's not used in many cases.
```python
if scrapy.priority:
    meta[b'scrapy_priority']=...
```

sibiryakov: @rolando what do you think? It looks fine. Even better than before, but I would used default priority.

rolando: Yes, sounds good to me. I will update the changes.

PR #262, opened 2017-02-22
WIP Allow spider to access frontier manager instance.

rolando:
For example, this helps to access or change frontera settings
dynamically.

### DIFF
--- a/frontera/contrib/scrapy/schedulers/frontier.py
+++ b/frontera/contrib/scrapy/schedulers/frontier.py
@@ -126,6 +126,8 @@ class FronteraScheduler(Scheduler):
 
     def open(self, spider):
         self.frontier.set_spider(spider)
+        if hasattr(spider, 'setup_frontier'):
+            spider.setup_frontier(self.frontier)
         self.logger.info("Starting frontier")
         if not self.frontier.manager.auto_start:
             self.frontier.start()


rolando:
Does it look good as a new feature? If so, I can add tests.

sibiryakov:
I would recommend to not change settings dynamically, because it makes debugging of your code terrible. What's the reason for accessing the settings from spider code?

sibiryakov:
@rolando should we move forward with this?

rolando:
> I would recommend to not change settings dynamically, because it makes debugging of your code terrible. What's the reason for accessing the settings from spider code?

The reason is to change frontera settings based on spider arguments. I haven't found a nicer way to do this.

sibiryakov:
@rolando could explain more, why is this needed?

rolando:
Specifically to set the HCF frontier slot via spider argument in Scrapy Cloud. Here is the code I use (having the patch applied):

```
class MySpider(Spider):

    # HCF frontier slot where to consume from.
    frontier_slot = None

    def setup_frontier(self, frontier):
        if self.frontier_slot is None:
            raise ValueError("frontier_slot not set")
        # HCFBackend is expected.
        frontier.manager.backend.consumer_slot = self.frontier_slot

```

Please let me know if there is a better way to do this.

rolando:
:thinking: possibly I could override the frontier slot in the frontera manager wrapper (method `set_spider`).

sibiryakov:
@rolando I think setting HCF frontier slot by means of settings is more natural then spider arguments. It's possible to pass per-job settings by means of REST API.

rolando:
I don't use Scrapy Cloud API, instead I rely on Periodic Jobs to ensure continuous running. I will see if I can find another workaround. Closing this PR for now.

sibiryakov:
@rolando let me know if you need any help with Periodic Jobs, if it's not possible to set Scrapy settings by means of periodic jobs then there is something wrong with them, and we could negotiate with SC team to add this feature.

PR #272, opened 2017-04-15
Fix error raising in msgpack and json decoders

ZipFile:
That's said. When upgrading from 0.7.0 to 0.7.1 my strategy/db workers went insane about not handled unknown message types coming from spiders.


sibiryakov:
Христос воскресе. Все посмотрю через неделю, отпуск.

A.

sibiryakov:
Thank you! 🍺

PR #274, opened 2017-04-20
Commit database cleared content

isra17:


frontera/contrib/backends/sqlalchemy/__init__.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, manager):
             session = self.session_cls()
             for name, table in DeclarativeBase.metadata.tables.items():
                 session.execute(table.delete())
+            session.commit()
```
sibiryakov: Are you sure that DDL statement needs commit?

isra17: At least on Postgres it does. Without this line, `SQLALCHEMYBACKEND_CLEAR_CONTENT` doesn't work.

sibiryakov: ok

sibiryakov:
Thank you!

PR #275, opened 2017-04-23
Unify backend logging

ZipFile:
Proposal: follow [python logging guides](https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial).

Resolution: reorganize code layout.

Also, since diff is already screwed up, fixed some flake8/pylint warnings.


ZipFile:
>The log length has exceeded the limit of 4 MB

:crying_cat_face:

frontera/contrib/backends/hbase/states.py
```diff
@@ -0,0 +1,80 @@
...
+from frontera.core.components import States
+from frontera.utils.misc import chunks
+
+from .utils import prepare_hbase_object
+
+
+logger = logging.getLogger(__name__)
+
+
+class HBaseState(States):
```
sibiryakov: This class describes states stored in hbase, so I think it should be plural: HBaseStates.

ZipFile: I tried to avoid any kind of renaming in order to keep it backward compatible.

ZipFile:
Check the changes now, hope this way is better.

sibiryakov:
@ZipFile I haven't forgot about this, just need some time to sort things out.

PR #276, opened 2017-04-25
Fix of requests propagation to Scheduler

sibiryakov:


sibiryakov:
Guys @ZipFile, @voith, @isra17 what do you think?

Now user can put spider middleware anywhere in the chain, but he has to mark requests as seeds if he wants them to be enqueued from Scrapy spider. Also there are clear rules of what passes to in-memory queue in scheduler.

I think that makes things clearer.

frontera/contrib/scrapy/schedulers/frontier.py
```diff
@@ -72,31 +72,43 @@ def _set_value(self, variable, value):
...
+        self._redirect_enabled = crawler.settings.get('REDIRECT_ENABLED')
         self._delay_next_call = 0.0
-        self.logger = getLogger('frontera.contrib.scrapy.schedulers.FronteraScheduler')
+        self.logger = getLogger('frontera.contrib.scrapy.schedulers.frontier.FronteraScheduler')
 
     @classmethod
     def from_crawler(cls, crawler):
         return cls(crawler)
 
     def enqueue_request(self, request):
```
isra17: Would it be even clearer to tag request that are enqueued in the Frontier through `process_spider_output -> links_extracted` (storing the tag in `meta` like you did for seed) and add any other non-tagged request to the local queue? This way you wouldn't need to check for redirect and other middleware could themselves chose between local scheduling or remote scheduling (Such as a middleware that logs in and try the request again without going through the frontier queue again).

ZipFile:
Quite a mess you have with merge requests, I think.

I'm not really sure about use case behind flagging requests, though. What kind of middleware you expect to be in the chain? What will be the source of seeds for them?

voith:
@sibiryakov Code wise Looks good to me. @isra17 has a good suggestion. I would like to know the pros and cons of your approach vs @isra17's suggestion.

frontera/contrib/scrapy/schedulers/frontier.py
sibiryakov: I believe adding middleware which is bypassing the frontier and want's to get it's requests in the local queue requires planning from the user operating the crawler. He would need to understand all the consequences of this. If we take your login example, the user would need to deal with response that comes after previously unseen login request. Frontera custom scheduler will crash on this, because it lacks of ``frontier_request`` in meta. Therefore the middleware which logs in would need to intercept this response or figure out something else.

Frontera already has some ancient mechanism of tagging the requests
https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/scrapy/converters.py#L41
at the moment it's not used.

sibiryakov: I see also other cases: redirects or website requires obtaining session. In any of these cases I expect user to look into Frontera custom scheduler code and make appropriate changes. I see such customization as advanced topic, requiring some expert knowledge.

sibiryakov:
@ZipFile 
> I'm not really sure about use case behind flagging requests, though. What kind of middleware you expect to be in the chain? What will be the source of seeds for them?

``enqueue_request()`` will get every request returned by spider and downloader middlewares and scheduled manually from Engine. The same is true for ``start_urls`` and ``start_requests`` fields in Spider class. So, there is no other way to figure out in scheduler ``enqueue_request`` method that request came from ``start_urls`` or ``start_requests`` except marking them explicitly in spider code.

Currently (master version) will transform every request coming from engine or middlewares to Frontera ``add_seeds`` event, which isn't right.

sibiryakov:
@ZipFile see above comment to Israel for suggestions of possible middlewares.

sibiryakov:
@redapple and @kmike, hi guys, we discuss integration between Frontera and Scrapy here. I believe this discussion could be interesting to you.

PR #277, opened 2017-04-27
Fix hostname local fingerprint from collisions

sibiryakov:
Unfortunately CRC32 is prone to collisions. According to my estimation about 0.001% of home page URLs in a seed list of 9M domains were lost because of current implementation of fingerprint. This PR attempts to solve that problem by adding hostname to the second part of which MD5 is calculated.


voith:
Hi @sibiryakov, This is comment has nothing to do with this PR but something on the line of fingerprint.
Frontera fingerprint doesn't work well for `POST` requests. Since fingerprint is  calculated only based on url and not the request body, this doesn't work well for `POST` requests.
Although I do understand that frontera is designed for broad crawling and `POST` requests won't be made for broad crawls, Does frontera ever plan to support POST requests?

I have a customised version of Fingerprint middleware that uses a customised version of [scrapys request fingerprint method](https://github.com/scrapy/scrapy/blob/master/scrapy/utils/request.py#L18-L62) to support POST requests

isra17:
For what it's worth, fingerprinting `POST` data is useful when crawling ASP application using hidden form to keep the application state.

sibiryakov:
Guys @voith and @isra17 I'm not against support of POST requests, please feel free to contribute! However POST requests are trickier.

I'll try to name some places were fingerprints are used:
- Links db keys, every link is uniquely identified using fingerprint,
- Keys for state management: cache in SW and states tables,
- Fingerprints are used in HBaseQueue for aggregations (but nothing major should happen),
- Message bus is using fingerprints to partition the requests across partitions in spider log and spider feed (if hostname partitioning is disabled).
- probably something else.

Actually we tried to incorporate form requests in our broad crawlers and those requests successfully  passed the queue and were fetched. Obviously they were overwriting each other in the link db, and were unusable in SW, but in general ``hostname_local_fingerprint`` worked for them. In SW I had to use custom data structures to maintain per-form states and link db was disabled.

PR #281, opened 2017-05-29
Mark partition as busy when a new batch is sent to it

isra17:
I happened to stumble onto this bug where the DBWorker keep sending request to the spider until the queue is empty. Seems to go as follow:

 2. DBWorker set Spider's partition as ready
 3. DBWorker send a new batch, `feed.counter = 256`
 4. Spider receive new batch, send new offset `spider.offset = 256`
 5. DBWorker receive offset, since `spider.offset <= feed.counter`, keep the partition as ready
 6. Spider is busy scraping.
 7. DBWorker send a new batch to the spider's partition, `feed.counter = 512`
 8. DBWorker still sending new batches, `feed.counter = 1024`
 9. Finally Spider has some space for new request, download next requests and then send its new offset, `spider.offset = 512`
10. DBWorker now set the partition as busy, however the lag between the spider offset and the feed counter can be quite huge by that time.

I guess crawling slowly make this worse since a single batch can take a few minutes to process, leaving the DBWorker some time to overload the feed.

My fix here is to mark a partition that received messages as busy. This way, the worker will wait for an update on the spider offset update to mark the partition as ready if needed. This should work well with a bigger `MAX_NEXT_REQUESTS` value on the worker to ensure the queue is never empty.

PS: Is there any IRC channel where the maintainers and other Frontera users are hanging out? I tried #scrapy but this didn't seem the right place to have Frontera discussion.

### DIFF
--- a/frontera/contrib/messagebus/zeromq/__init__.py
+++ b/frontera/contrib/messagebus/zeromq/__init__.py
@@ -8,7 +8,7 @@ import zmq
 import six
 
 from frontera.core.messagebus import BaseMessageBus, BaseSpiderLogStream, BaseStreamConsumer, \
-    BaseSpiderFeedStream, BaseScoringLogStream
+    BaseSpiderFeedStream, BaseScoringLogStream, BaseStreamProducer
 from frontera.contrib.backends.partitioners import FingerprintPartitioner, Crc32NamePartitioner
 from frontera.contrib.messagebus.zeromq.socket_config import SocketConfig
 from six.moves import range
@@ -61,7 +61,7 @@ class Consumer(BaseStreamConsumer):
         return self.counter
 
 
-class Producer(object):
+class Producer(BaseStreamProducer):
     def __init__(self, context, location, identity):
         self.identity = identity
         self.sender = context.zeromq.socket(zmq.PUB)
@@ -80,7 +80,7 @@ class Producer(object):
         # Raise TypeError if any message is not encoded as bytes
         if any(not isinstance(m, six.binary_type) for m in messages):
             raise TypeError("all produce message payloads must be type bytes")
-        partition = self.partitioner.partition(key)
+        partition = self.partition(key)
         counter = self.counters.get(partition, 0)
         for msg in messages:
             self.sender.send_multipart([self.identity + pack(">B", partition), msg,
@@ -99,6 +99,9 @@ class Producer(object):
 
     def get_offset(self, partition_id):
         return self.counters.get(partition_id, None)
+
+    def partition(self, key):
+        return self.partitioner.partition(key)
 
 
 class SpiderLogProducer(Producer):
--- a/tests/mocks/message_bus.py
+++ b/tests/mocks/message_bus.py
@@ -1,5 +1,5 @@
 from frontera.core.messagebus import BaseMessageBus, BaseSpiderLogStream, BaseStreamConsumer, \
-    BaseScoringLogStream, BaseSpiderFeedStream
+    BaseScoringLogStream, BaseSpiderFeedStream, BaseStreamProducer
 
 
 class Consumer(BaseStreamConsumer):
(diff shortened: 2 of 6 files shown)


sibiryakov:
@isra17 thanks for the contribution!
we don't have irc or other chat channel, because there is not that big demand.

I'm not sure I understand what the problem is: 
- loose of some requests generated by DBW between 7 and 8 steps? 
or
- incorrect partition status set because of wrong sequence/timing of offset exchange?

A.

isra17:
The issue is that until the Spider finishes its current batch, the DBWorker will just keep sending new ones. In my case, the DBWorker have time to flush the entire backend queue into the message bus before the spider has the opportunity to mark itself as busy. This get annoying when the spider ends up with a few hours worth of work waiting in the message bus.

sibiryakov:
> The issue is that until the Spider finishes its current batch, the DBWorker will just keep sending new ones. In my case, the DBWorker have time to flush the entire backend queue into the message bus before the spider has the opportunity to mark itself as busy.
The idea behind the code you're trying to modify is that DBW sends always some amount of request in advance. When there's a pauses between batches required for passing the states (ready->busy->ready) and waiting for a batch to finish (when batch is 95% finishing the spider is mostly idle waiting for a longest request) the crawling speed decreases. With some amount of requests always available in the queue spider has a chance to get requests always when there's a space in it's internal queue.

> This get annoying when the spider ends up with a few hours worth of work waiting in the message bus.

I don't understand this. Spider is waiting because a) messages with batches were lost in ZMQ or b) busy status was incorrectly set and wasn't changing for long time, even when spider was already ready?

This is pretty tough topic to discuss async/remotely, so please contact me by Skype,  alexander.sibiryakov so we could save some time.


A.

isra17:
@sibiryakov I did refactor the PR to keep track of the offset as discussed on Skype. Let me know if anything is missing.

tests/mocks/message_bus.py
```diff
@@ -70,23 +73,28 @@ def consumer(self, partition_id, type):
...
-        return self.ready_partitions
-
-    def mark_ready(self, partition_id):
-        self.ready_partitions.add(partition_id)
-
-    def mark_busy(self, partition_id):
-        self.ready_partitions.discard(partition_id)
-
+        partitions = []
+        for partition_id, last_offset in self.partitions_offset.items():
```
sibiryakov: Here, if partition doesn't exist (yet) in this dict - it will not be returned as available, which is wrong. 
What if producer offsets will be less (because of DB worker restart) than consumer offsets?

frontera/contrib/messagebus/zeromq/__init__.py
```diff
@@ -176,22 +176,32 @@ def __init__(self, messagebus):
...
+        for partition_id, last_offset in self.partitions_offset.items():
+            lag = self._producer.get_offset(partition_id) - last_offset
+            if lag < self.max_next_requests:
+                partitions.append(partition_id)
+        return partitions
 
-    def mark_busy(self, partition_id):
-        self.ready_partitions.discard(partition_id)
+    def set_spider_offset(self, partition_id, offset):
+        self.partitions_offset[partition_id] = offset
```
isra17: Ha good catch

tests/mocks/message_bus.py
isra17: line 86 does create the keys for each partition. In the worst case, a new partition will first send an `offset` message and create a key in the `partitions_offset`. As for the negative offset I don't think anything will break? From my understanding, when a DBWorker restart, a spider will have a big invalid offset, but it should be marked as ready and on its next message [this will fix it](https://github.com/scrapinghub/frontera/blob/master/frontera/contrib/messagebus/zeromq/__init__.py#L49).

sibiryakov: Ok!

PR #286, opened 2017-07-18
Use CityHash for FingerprintPartitioner

vshlapakov:
Google's CityHash provides better performace, the PR uses [its python bindings](https://github.com/escherba/python-cityhash) to improve current fingerprint partitioner logic. My only worries - if it should be a recommended way and a required dependency, or one of the package extras.


sibiryakov:
merged in #299

PR #295, opened 2017-10-05
Set overused factor so it actually trigger when a slot is full.

isra17:
The settings is used to compare against: `overused_factor = len(slot.active) / float(slot.concurrency)`.
Unless I missed something, overused_factor is going to be between `[0..1]` and the following condition `overused_factor > self.frontier.manager.settings.get('OVERUSED_SLOT_FACTOR')` is never going to be `True` with a value of `5.0`. Setting a value of `1.0` at least start buffering when a slot is full.


isra17:
nvm, got it wrong!

PR #296, opened 2017-10-06
Add max_queue_size to overused buffer to stop queue request when the …

isra17:
…pending queue hit a size.

This prevent a single spider to buffer too many requests. Deactivated by default, must be used with backends that support spiders dropping requests.


sibiryakov:
Hi @isra17 sorry for a delay, have a look here
https://github.com/scrapinghub/frontera/pull/299

there is a slightly more advanced solution implemented.

sibiryakov:
closing it because of duplicated functionality in #299

PR #306, opened 2017-12-27
Update hbase.py

clarksun:
imho, in hbase queue's `get_next_requests` method,  params: min_hosts and max_requests_per_host are optional, because this method will check whether they are None, eg line no.225 and no.236, but kwargs.pop will raise Exception if I don't give these two params.


sibiryakov:
thank you @clarksun

PR #308, opened 2017-12-27
Update db.py

clarksun:
if dbw is a `batch server` or `score log process server`, I mean this db process do not consume spider log topic, After add `no-incoming` flag to db launch command,  group `dbw-spider-log` always log 'Heartbeat session expired. so I add this condition to prevent init spider_log.consumer

### DIFF
--- a/frontera/worker/db.py
+++ b/frontera/worker/db.py
@@ -76,7 +76,8 @@ class DBWorker(object):
         spider_log = self.mb.spider_log()
 
         self.spider_feed = self.mb.spider_feed()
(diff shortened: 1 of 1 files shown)


sibiryakov:
Hi, have you tested it @clarksun ? Is DBW with --no-incoming option working?

clarksun:
@sibiryakov this is our dbw. https://gist.github.com/clarksun/f651c2200164ce9999428d858e67d81b

batch server and scoring server both add --no-incoming, and works well.

Prometheus3375:
Seems like this PR is outdated, so it can be closed.

PR #310, opened 2017-12-28
HBaseState flush when frontier stop

clarksun:
flush method in HBaseState invoked every 5 minutes by default settings to save cache state into meta table, add frontier_stop method to prevent state loss from memory cache dict when sw instance terminated.


sibiryakov:
Good catch @clarksun! I'm going to merge it.

sibiryakov:
I have checked this code again, @clarksun and found that SW was already running the flush when stopping https://github.com/scrapinghub/frontera/blob/master/frontera/worker/strategy.py#L291, if we apply your patch it will be flushing two times. I propose to remove the state cache flush on stopping from SW then and make sure `frontier_stop` will be executed.

PR #317, opened 2018-01-30
Add support for “post requests and data” when integrating scrapy.

CN-hanyi:
Add support for “post requests and data” when integrating scrapy.
Resolve the bug for “wrong request data in metadata” when response updates metadata.

### DIFF
--- a/frontera/contrib/backends/remote/codecs/msgpack.py
+++ b/frontera/contrib/backends/remote/codecs/msgpack.py
@@ -32,12 +32,12 @@ def _prepare_request_message(request):
         else:
             logger.warning('unable to serialize object: {}'.format(obj))
             return None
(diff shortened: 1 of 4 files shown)


sibiryakov:
Hi @CN-hanyi thank you for your PR! Would be great to have a tests passing and to propagate your changes to all backends: Memory, HBase, MessageBus. Probably it's also worth to check if it needs to be mentioned in docs.

CN-hanyi:
Thanks for your reply. I have added the test for the submitted code.
Add test for “post request and data support when integrating scrapy ” ,
propagate this change to all backends: memory, hbase,redis .
propagate this change to json codecs。
Add MessageBusBackend test for new codecs(json and msgpack)。
Fix some bugs after the test.

PR #329, opened 2018-05-08
Framed transport for Kafka message bus

sibiryakov:
Allows to transfer messages larger than Kafka maximum message size.

### DIFF
--- a/frontera/contrib/messagebus/kafkabus.py
+++ b/frontera/contrib/messagebus/kafkabus.py
@@ -14,11 +14,16 @@ from frontera.core.messagebus import BaseMessageBus, BaseSpiderLogStream, BaseSp
 from twisted.internet.task import LoopingCall
 from traceback import format_tb
 from os.path import join as os_path_join
+from msgpack import Packer, unpackb
+from collections import defaultdict
+from math import ceil
+from cachetools import LRUCache
 
 
 DEFAULT_BATCH_SIZE = 1024 * 1024
 DEFAULT_BUFFER_MEMORY = 130 * 1024 * 1024
 DEFAULT_MAX_REQUEST_SIZE = 4 * 1024 * 1024
+MAX_SEGMENT_SIZE = int(DEFAULT_MAX_REQUEST_SIZE * 0.95)
 
 logger = getLogger("messagebus.kafka")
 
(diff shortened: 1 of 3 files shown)


jpbalarini:
Hi @sibiryakov, is there any reason this was not merged yet? I'm dealing with a case where this should fix my problem.

Thanks!

frontera/contrib/messagebus/kafka/transport.py
```diff
@@ -0,0 +1,51 @@
...
+import hashlib
+from cachetools import LRUCache
+from msgpack import Packer, unpackb
+from random import randint
+from six import MAXSIZE
+from struct import pack
+
+
+def random_bytes():
+    return pack("L", randint(0, MAXSIZE))
```
Prometheus3375, line 12: You have an error here. Look at `six` sources

```python
if PY3:
    string_types = str,
    integer_types = int,
    class_types = type,
    text_type = str
    binary_type = bytes

    MAXSIZE = sys.maxsize
else:
    string_types = basestring,
    integer_types = (int, long)
    class_types = (type, types.ClassType)
    text_type = unicode
    binary_type = str

    if sys.platform.startswith("java"):
        # Jython always uses 32 bits.
        MAXSIZE = int((1 << 31) - 1)
    else:
        # It's possible to have sizeof(long) != sizeof(Py_ssize_t).
        class X(object):

            def __len__(self):
                return 1 << 31
        try:
            len(X())
        except OverflowError:
            # 32-bit
            MAXSIZE = int((1 << 31) - 1)
        else:
            # 64-bit
            MAXSIZE = int((1 << 63) - 1)
        del X
```

`sys.maxsize` is 64 bit in Python 3 (9223372036854775807). As you can see from six, MAXSIZE can be 64 bit both for Python 2 and 3. According to documentation (Python [2](https://docs.python.org/2.7/library/struct.html#format-characters) and [3](https://docs.python.org/3/library/struct.html#format-characters)), packing symbol `L` is used to pack 32 bit uint. You should change it to `Q` which is used to pack 64 bit uint.

```suggestion
def random_bytes():
    return pack('Q', randint(0, MAXSIZE))
```

Prometheus3375: Another option is to change `MAXSIZE` variable to literal `4_294_967_295`.

PR #352, opened 2018-11-15
Fix bug when ordering elements on queue when get next batch of requests

jpbalarini:
I found that when using a MySQL database as backend the `queue` table was being consumed in the opposite order as intended (elements with lower priority were being consumed first).

I found that on the `queue` component the `_order_by` method had no default order condition, so I added an explicit `desc()` clause (elements with bigger priority first). I guess this can apply to other backends (Postgres).

Thanks!

### DIFF
--- a/frontera/contrib/backends/sqlalchemy/components.py
+++ b/frontera/contrib/backends/sqlalchemy/components.py
@@ -164,7 +164,7 @@ class Queue(BaseQueue):
             return query.order_by(self.queue_model.created_at)
         if self.ordering == 'created_desc':
             return query.order_by(self.queue_model.created_at.desc())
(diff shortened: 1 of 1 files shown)


icapurro:
Great! Thanks for this fix!

sibiryakov:
hey @jpbalarini thanks for the contribution, will try to respond quicker next time. It seems this wasn't covered with the test, can you add them?

jpbalarini:
@sibiryakov done! added tests for the SQL alchemy `Queue` component. That class should have 100% code coverage now.

Thanks!

jpbalarini:
Hi @sibiryakov I added the tests and everything as requested some time ago.
Thanks!

Prometheus3375:
When this fix will be applied?

PR #353, opened 2018-11-15
Fix sqlalchemy queue component partition_id

jpbalarini:
I found that the partition id was always set to 0 when disabling the `QUEUE_HOSTNAME_PARTITIONING` setting, using a sqlalchemy backend. The problem was that the `schedule` method on the `Queue` component was always using the `hostname` as the partitioning key (independent of the QUEUE_HOSTNAME_PARTITIONING setting).

Thanks!


icapurro:
Thanks for this fix @jpbalarini !

jpbalarini:
@sibiryakov what do you think?
Thanks!

jpbalarini:
@sibiryakov what do you think of these changes? Adding them will fix the case when someone disables hostname partitioning (my case)

PR #357, opened 2018-11-29
use 2.1.0 version of cachetools

vezunch1k:


sibiryakov:
already updated to support the latest version

PR #361, opened 2019-01-21
Update after redis update

khellan:
Use dict of scheduled items rather than a list after redis library update.


sibiryakov:
Ok, looks good! Thank you 👍

PR #364, opened 2019-02-24
Add 3.6 support in setup.py

lucywang000:
hi, I'm trying to learn about frontera and I hope this can be my a starting point of small contribution.

Currently in setup.py only python 3.5 is claim to be supported, I think we can add 3.6 too.

There are a few tests errors in travis ci, and I don't intent to fix them here - maybe for another pr - imo as long as there are same number of fails for python 3.5 and 3.6 then we're good.

Errors for py35 (build log https://travis-ci.com/lucywang000/frontera/jobs/180100043)
errors-py35

Errors for py36(build log https://travis-ci.com/lucywang000/frontera/jobs/180100044)
errors-py36

For 3.7 there is a problem: thriftpy can't install under 3.7, and that project is no longer maintained:
``` 
 thriftpy/transport/cybase.c:3192:11: error: ‘PyThreadState {aka struct _ts}’ has no member named ‘exc_traceback’
       tstate->exc_traceback = tb;
             ^
  error: command 'gcc' failed with exit status 1
  
  ----------------------------------------
  Failed building wheel for thriftpy
```

but there is a new fork called thriftpy2 (see https://github.com/Yelp/py_zipkin/pull/115). This may also be the work of a separate pr.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,14 @@ matrix:
         - mysql
         - postgresql
         - redis-server
+    - python: 3.7
+      env: TOXENV=py37
+      dist: xenial
+      services:
+        - docker
+        - mysql
+        - postgresql
+        - redis-server
 
 install:
   - pip install -U tox wheel codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,14 +35,6 @@ matrix:
         - mysql
         - postgresql
         - redis-server
-    - python: 3.7
-      env: TOXENV=py37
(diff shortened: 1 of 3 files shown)


sibiryakov:
Hi @lucywang000 it looks good, but I think we should fix the test before merging this, there could be issues connected with Python3.6 but we're unable to find them unless tests are fixed.

suzil:
This PR can be closed as there is now 3.6 + 3.7 support.

PR #379, opened 2019-10-23
Bump to Python 3.6 + 3.7

suzil:
- [thriftpy](https://github.com/Thriftpy/thriftpy) is no longer maintained; [thriftpy2](https://github.com/Thriftpy/thriftpy2) is the active implementation
- `test_zmq_message_bus()` sometimes fails, so I made the test [flaky](https://pypi.org/project/flaky/) so that it will try to rerun the test one time if the first try fails

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
         'Topic :: Internet :: WWW/HTTP',
         'Topic :: Software Development :: Libraries :: Application Frameworks',
         'Topic :: Software Development :: Libraries :: Python Modules',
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -34,14 +34,17 @@
...
         'Topic :: Internet :: WWW/HTTP',
         'Topic :: Software Development :: Libraries :: Application Frameworks',
         'Topic :: Software Development :: Libraries :: Python Modules',
     ],
     install_requires=[
         'six>=1.8.0',
         'w3lib>=1.15.0',
-        'cityhash>=0.1.7'
+        'cityhash>=0.1.7',
+        'thriftpy2'
```
sibiryakov: the same with thriftpy

suzil:
thriftpy2 is fully compatible with thriftpy according to their docs (and they are in the same organization), so it should be OK. I have included thriftpy2 additionally as a requirement of hbase.

sibiryakov:
okay! merging

PR #380, opened 2019-10-24
Use absolute imports

suzil:
* Absolute imports are easier to read and allow files to be moved around without breaking them

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,6 +99,7 @@ setup(
         "colorlog>=2.4.0",
         "python-json-logger>=0.1.5",
         "redis>=2.10.5",
-        "hiredis>=0.2"
+        "hiredis>=0.2",
+        "flake8-absolute-import",
     ]
 )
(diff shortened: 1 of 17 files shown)


setup.py
```diff
@@ -41,7 +41,8 @@
     install_requires=[
         'six>=1.8.0',
         'w3lib>=1.15.0',
-        'cityhash>=0.1.7'
+        'cityhash>=0.1.7',
+        "thriftpy"
```
sibiryakov: thriftpy is only needed when installed with hbase and distributed stuff

suzil: Tests broke when I didn't include it, should it be in a different section for installs?

sibiryakov: and yes, thtiftpy2

sibiryakov:
Conflict @suzil

setup.py
suzil: Fixed

suzil:
Fixed now @sibiryakov

suzil:
closing due to no response

PR #383, opened 2019-10-25
Move tox to test requirements

suzil:
- Sorted the test requirements to make it a bit easier to read
- Since we already have a `setup.cfg` file, move `.coveragerc` contents to it. I think since the `setup.cfg` file is there, coverage might've not even been reading the `.coveragerc` file, but I'm not sure. Either way, it's good to combine the files and move CI config settings all to one place

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,0 @@
-[run]
-branch = true
-source = frontera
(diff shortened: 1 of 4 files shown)


suzil:
closing as no response

PR #384, opened 2019-10-25
Add "make run" command

suzil:
Makes for easier local development :slightly_smiling_face: 

- Add a `make run` command for easier local development
- Move the test docker-compose.yml file to the root directory (this is more standard)
- Bump docker-compose config version to 3 (latest is best if it works!)
- Remove installing a separate docker-compose version as it seems like the version included by Travis works for running tests
- Add note to README on new Makefile command for local development

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,6 @@ services:
   - redis-server
 
 env:
-  global:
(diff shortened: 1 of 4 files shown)


suzil:
closing as no response

PR #386, opened 2019-10-25
Add a test for generating graph visualizations

suzil:


suzil:
closing due to no response

PR #387, opened 2019-10-26
Enable test_scrapy_spider() test as it seems to be working now

suzil:


suzil:
It seems this is due to starting multiple reactors (one in test_scrapy_spider() and another in StrategyWorker). I think the proper solution here would be to use something like a [CrawlerProcess](https://docs.scrapy.org/en/latest/topics/api.html#scrapy.crawler.CrawlerProcess) and to not execute reactor.run() inside the main code but instead either let the user run it or to use the wrapper like CrawlerProcess. I'm not sure though

PR #391, opened 2019-12-23
fixed incorrect import module

cielpy:


Gallaecio:
I think this should be accompanied by a change to `setup.py` to require `Scrapy>=1.0.0`, since [Scrapy 1.0.0 renamed that module](https://docs.scrapy.org/en/latest/news.html#scrapy-1-0-0-2015-06-19).

